Issue #119, opened 2019-04-04
`sudo bin/setup_base` does not install openvswitch on Ubuntu 18.04.2 LTS


InBrewJ:
Running `sudo bin/setup_base` results in the following output:

> AG is sudo apt-get -qqy --no-install-recommends
E: Package 'realpath' has no installation candidate
Installing for ubuntu bionic
deb http://packages.wand.net.nz bionic main
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  2221  100  2221    0     0   4105      0 --:--:-- --:--:-- --:--:--  4105
OK
Hit:1 http://ppa.launchpad.net/videolan/master-daily/ubuntu bionic InRelease
Hit:2 http://gb.archive.ubuntu.com/ubuntu bionic InRelease                        
Get:3 http://security.ubuntu.com/ubuntu bionic-security InRelease [88.7 kB]       
Ign:4 http://dl.google.com/linux/chrome/deb stable InRelease                      
Hit:5 http://gb.archive.ubuntu.com/ubuntu bionic-updates InRelease                
Hit:6 http://packages.microsoft.com/repos/vscode stable InRelease                 
Hit:7 https://download.docker.com/linux/ubuntu bionic InRelease                   
Hit:8 http://ppa.launchpad.net/wireshark-dev/stable/ubuntu bionic InRelease       
Get:9 http://gb.archive.ubuntu.com/ubuntu bionic-backports InRelease [74.6 kB]    
Hit:10 https://dl.yarnpkg.com/debian stable InRelease                             
Hit:11 http://dl.google.com/linux/chrome/deb stable Release                       
Hit:12 http://packages.cloud.google.com/apt cloud-sdk-bionic InRelease            
Hit:13 http://packages.wand.net.nz bionic InRelease                               
Fetched 163 kB in 1s (185 kB/s)                    
Reading package lists... Done
E: Packages were downgraded and -y was used without --allow-downgrades.

as a result, `sudo apt-get install openvswitch-switch openvswitch-common` is never run and when tests are executed daq fails.

Workaround: run `sudo apt-get install openvswitch-switch openvswitch-common` manually

grafnu:
[...]

PR #147, opened 2019-06-18
Setup base downgrade bug

jhughesbiot:
Allow more control over apt-get arguments to correct issue #119

### DIFF
--- a/bin/setup_base
+++ b/bin/setup_base
@@ -10,10 +10,19 @@
 DOCKER_VERSION_BASE="=18.06.1~ce~3-0"
 
 if [ -z "$AG" ]; then
-    AG="sudo apt-get -qqy --no-install-recommends"
+    AG="sudo apt-get -q --no-install-recommends"
 fi
 
-echo AG is $AG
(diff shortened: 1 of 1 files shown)


grafnu:
Unfortunately it looks like this breaks the Travis build -- as per
https://travis-ci.org/grafnu/daq/jobs/547484211

The problem is the pinned version of docker -- which was there to work
around some bug (as per comments).  But, not sure what the right fix is --
but whatever it is likely requires a fair bit of testing to get right.
Build dependencies are very annoying :-/

I'm not sure if the docker workaround/hack is still required...

jhughesbiot:
I can put the original apt-get parameters back in 

> -qqy

 and leave the additional arguments there.  This would preserve it's current behavior and would should through the CI as before.  This would just allow people encountering the issue of adding the optional  --allow-downgrades for local builds for people with a higher docker version already installed.

grafnu:
Can you pls. submit against grafnu/daq?  That has the proper CI testing in place!

PR #165, opened 2019-07-19
reordered file access to gcp_cred related values in test_udmi, pubber…

InBrewJ:
… now doesn't start if there is no pubber.json - this should make the udmi test skip gracefully if there is no GCP_SERVICE_ACCOUNT in a given Travis.org account

### DIFF
--- a/subset/cloud/test_udmi
+++ b/subset/cloud/test_udmi
@@ -6,7 +6,7 @@ REPORT=/tmp/report.txt
 route add default gw $GATEWAY_IP
 
 gcp_cred=/config/inst/gcp_service_account.json
-gcp_topic=target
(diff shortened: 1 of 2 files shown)


subset/cloud/test_udmi
```diff
@@ -6,7 +6,7 @@ REPORT=/tmp/report.txt
 route add default gw $GATEWAY_IP
 
 gcp_cred=/config/inst/gcp_service_account.json
-gcp_topic=target
+gcp_topic=events
```
InBrewJ: https://github.com/faucetsdn/daq/blob/master/pubber/src/main/java/daq/pubber/Pubber.java#L32

PR #219, opened 2020-01-15
Adding timeout for docker tests with default timeout

henry54809:
And handle scenarios where DHCP assignments happen before gateway is ready for activation.


daq/dhcp_monitor.py
```diff
@@ -97,5 +97,5 @@ def _dhcp_error(self, e):
         LOGGER.error('DHCP monitor %s error: %s', self.name, e)
         if self.dhcp_log:
             self.dhcp_log.write('Monitor error %s\n' % e)
-        self.cleanup(forget=False)
         self.callback('error', None, exception=e)
+        self.cleanup()
```
grafnu: Are you sure this works in the case where there's a legit DHCP error?  E.g., if the stream is closed first, causing the error?  The purpose of the (forget=False) argument, IIRC, is to handle the case where stream closes, error happens, and then the stream shouldn't be closed again (which would be an error).  Anyway, as long as it works it's all good -- but please check that case.  You can do this by manually (externally to DAQ) killing the DHCP monitor process and see what happens.

daq/host.py
```diff
@@ -61,6 +61,7 @@ class ConnectedHost:
     """Class managing a device-under-test"""
 
     _MONITOR_SCAN_SEC = 30
+    _DEFAULT_TIMEOUT = 350
```
grafnu: we'll need a way to apply a global setting to this without changing code.  See _MONITOR_SCAN_SEC and how it pulls a value from config, and then add a similar commented-out example to misc/system.conf (search for monitor_scan_sec). Otherwise, we run into the situation where we might break existing installs, which is OK if there's an easy way to recover (e.g. adding a global config).

daq/host.py
```diff
@@ -247,6 +253,12 @@ def _prepare(self):
         self._state_transition(_STATE.WAITING, _STATE.INIT)
         self.record_result('sanity', state=MODE.DONE)
         self.record_result('dhcp', state=MODE.EXEC)
+        _ = list(listener(self) for listener in self._dhcp_listeners)
```
grafnu: comprehension [] preferred over list()   (just more pythonic)

daq/host.py
```diff
@@ -452,7 +464,8 @@ def _docker_test(self, test_name):
         }
 
         self.test_host = docker_test.DockerTest(self.runner, self.target_port, self.devdir,
-                                                test_name)
+                                                test_name,
+                                                timeout_sec=self._get_test_timeout(test_name))
```
grafnu: don't include function-call inline (just harder to read), extract to local variable and then can combine on same line as test_name -- same number of overall lines but easier to read.

daq/host.py
```diff
@@ -247,6 +253,12 @@ def _prepare(self):
         self._state_transition(_STATE.WAITING, _STATE.INIT)
         self.record_result('sanity', state=MODE.DONE)
         self.record_result('dhcp', state=MODE.EXEC)
+        _ = list(listener(self) for listener in self._dhcp_listeners)
+
+    def register_dhcp_ready_listener(self, callback):
+        """Registers callback for when the host is ready for activation"""
+        if callback:
```
grafnu: when is callback not defined -- is there ever a valid use-case for that?  Seems like it would be better to error out in that case or pass it along.  Silently ignoring errors usually isn't a good thing.

daq/host.py
```diff
@@ -147,6 +149,10 @@ def _test_enabled(self, test):
         test_module = self._loaded_config['modules'].get(test)
         return test in self._CORE_TESTS or test_module and test_module.get('enabled', True)
 
+    def _get_test_timeout(self, test):
+        test_module = self._loaded_config['modules'].get(test)
+        return test_module.get('timeout', self._DEFAULT_TIMEOUT) if test_module else None
```
grafnu: what are the cases where test is not defined in modules?  Seems like that might be the implicit modules (pass/fail/etc...)... just want to make sure the logic here is correct. Alternative, of course, would be to apply the default timeout in that case as well. Is there a reason not to do that?

The only one I can think of is the "test_hold" module, which is explicitly design/intended to block/hold forever.

daq/stream_monitor.py
```diff
@@ -4,35 +4,40 @@
...
         """Start monitoring a specific descriptor"""
         fd = self.get_fd(desc)
         assert fd not in self.callbacks, 'Duplicate descriptor fd %d' % fd
         if copy_to:
             assert not callback, 'Both callback and copy_to set'
             self.make_nonblock(desc)
             callback = lambda: self.copy_data(name, desc, copy_to)
-        LOGGER.debug('Monitoring start %s fd %d', name, fd)
-        self.callbacks[fd] = (name, callback, hangup, error)
+        LOGGER.debug('Monitoring start %s fd %d', name, fd, )
```
grafnu: extra comma?

daq/stream_monitor.py
```diff
@@ -4,35 +4,40 @@
...
+        self.callbacks[fd] = (name, callback, hangup, error,
+                              datetime.fromtimestamp(time.time())
```
grafnu: extract to local varaible -- inline logic is messy

daq/stream_monitor.py
```diff
@@ -145,16 +149,18 @@ def event_loop(self):
...
                 for fd, event in fds:
-                    self.process_poll_result(event, fd)
-                    if self.dirty:
-                        LOGGER.debug('Monitoring set modified, re-issuing poll...')
-                        break
-            else:
-                LOGGER.debug('Monitoring no fds found')
-                return True
+                    if fd in self.callbacks:
+                        self.process_poll_result(event, fd) #Monitoring set could be modified
```
grafnu: space after #, also move to previous line to indicate *why* the callback list is checked.  (e.g., could have been modified so the callbacks are no longer valid).

daq/stream_monitor.py
```diff
@@ -145,16 +149,18 @@ def event_loop(self):
...
+            # check for timeouts
+            frozen_callbacks = copy.copy(self.callbacks)
+            for fd in frozen_callbacks:
+                name, _, _, on_error, timeout = frozen_callbacks[fd]
+                if timeout and datetime.fromtimestamp(time.time()) >= timeout:
+                    self.error_handler(fd, TimeoutError("Timed out."), name, on_error)
+                    if fd in self.callbacks:
+                        self.forget(fd)
```
grafnu: This seems indeterminate here -- in that error_handler *should* or must remove the callback, otherwise it's a logical error. I would go the other way... assert that the fd is in fact removed from the callbacks and throw an error if not. -- unless there is some fundamental reason why that doesn't happen.

testing/test_aux.out
```diff
@@ -65,10 +65,12 @@ port-01 module_config modules
     "enabled": true
   },
   "nmap": {
-    "enabled": true
+    "enabled": true,
+    "timeout": 60
```
grafnu: I don't see anything where you actually test the timeout capability?  There should be something in here that is expected to timeout (e.g. you could go the other way and set a very short timeout on one of these enabled tests), and then explicitly check for that expected failure in the output results.  E.g., if the timeout code gets messed up for some reason, how would we know (which is the purpose of the integration tests).

henry54809:
PTAL. travis build was successful.

grafnu:
I put together a "reverse PR" ( <https://github.com/henry54809/daq/pull/4>if
that's what you call it) against your repo that removes the try/except
blocks and uses explicit logic instead.   You should be able to merge that
into your branch, make sure it's still WAI, and then should be good to go.

henry54809:
I made changes to clean up the stream shutdown logic and removed the try, except that was in place. PTAL.

testing/test_aux.out
```diff
@@ -40,7 +40,7 @@ RESULT skip security.firmware Could not retrieve a firmware version with nmap. C
 RESULT pass security.firmware version found: ?\xFF\xFF\x19,>u\x08\x00no
 dhcp requests 2 1
 01: ['01:brute:1', '01:macoui:1']
-02: ['02:nmap:1', '02:tls:1']
+02: ['02:hold:Timed out.', '02:nmap:1', '02:tls:1']
```
grafnu: needs to be a programatic single identified (e.g. "timeout"), not a string with spaces. Design intent is that it can be used for things like CSS or whatever, so want to keep it simple.

daq/stream_monitor.py
```diff
@@ -97,6 +102,7 @@ def trigger_hangup(self, fd, event):
         name = self.callbacks[fd][0]
         callback = self.callbacks[fd][2]
         on_error = self.callbacks[fd][3]
+        self.callbacks[fd][5].close() # close the stream
```
grafnu: redundant comment

daq/stream_monitor.py
```diff
@@ -32,7 +35,9 @@ def monitor(self, name, desc, callback=None, hangup=None, copy_to=None, error=No
             self.make_nonblock(desc)
             callback = lambda: self.copy_data(name, desc, copy_to)
         LOGGER.debug('Monitoring start %s fd %d', name, fd)
-        self.callbacks[fd] = (name, callback, hangup, error)
+        timeout = datetime.fromtimestamp(time.time()) + \
```
grafnu: extract second clause to variable -- same number of lines and avoids line continuation

delta = timedelta(seconds=timeout_sec) if timeout_sec else None
timeout = datetime.fromtimestamp(time.time()) + delta

If that's the right logic -- that's another reason -- the precedence between "+" and "if" is not obvious.

misc/module_config.json
```diff
@@ -1,10 +1,12 @@
 {
   "modules": {
     "pass": {
-      "enabled": true
+      "enabled": true, 
+      "timeout": 60
```
grafnu: yah, I think these should be timeout_sec to avoid confusion. Also the default is default_timeout_sec so it makes it consistent.

grafnu: What does this (time out on the "pass" module) do?  Pass should pretty much complete instantly?

misc/module_config.json
```diff
@@ -1,10 +1,12 @@
...
     },
     "nmap": {
-      "enabled": true
+      "enabled": true,
+      "timeout": 60
```
grafnu: why make this change?  When I checked nmap was pretty quick (seconds) -- is there a case where 60 fails?

daq/stream_monitor.py
```diff
@@ -149,16 +153,18 @@ def event_loop(self):
...
-                LOGGER.debug('Monitoring no fds found')
-                return True
+                    if fd in self.callbacks: # Make sure it's still there.
+                        self.process_poll_result(event, fd)
+            # Check for timeouts
+            frozen_callbacks = copy.copy(self.callbacks)
+            for fd in frozen_callbacks:
+                name, _, _, on_error, timeout, _ = frozen_callbacks[fd]
+                if timeout and datetime.fromtimestamp(time.time()) >= timeout:
+                    LOGGER.error('Monitoring timeout fd %d (%s)', fd, name)
```
grafnu: add the actual timeout value to the logging message, I think that will be useful for diagnosing any problems

testing/test_aux.out
```diff
@@ -101,16 +103,18 @@ port-02 module_config modules
...
-    "enabled": false
+    "enabled": true 
   },
   "macoui": {
     "enabled": true
   },
   "nmap": {
-    "enabled": true
+    "enabled": true,
+    "timeout": 60
```
grafnu: was there a case somewhre that was testing that the JSON paramater was actually doing something. I'm not sure what is currently being tested with the hold timeout, but it might be the default (350?).   Not sure how to easily verify that it works, but would like to at least be able to tell from the logs.

daq/host.py
```diff
@@ -344,20 +359,20 @@ def _base_start(self):
             self._monitor_cleanup()
             self._monitor_error(e)
 
-    def _monitor_cleanup(self, forget=True):
+    def _monitor_cleanup(self):
         if self._tcp_monitor:
             LOGGER.info('Target port %d monitor scan complete', self.target_port)
-            if forget:
+            if self._tcp_monitor.stream() and not self._tcp_monitor.stream().closed:
```
grafnu: I'm still not a fan of these explicit "close if not closed" checks... I would want to know the case when we don't know in advance that the stream is closed already (it can happen, but should be the exception corner-case not the default behavior). I think there's this one and then one in gateway.py.  This was the reason behind the forget argument.  Anyway -- I took them out and I'm running the tests again, so we can see what happens...

daq/stream_monitor.py
henry54809: I think I'll use traditional if statement to make it clearer.

daq/host.py
grafnu: Ok, created https://github.com/henry54809/daq/pull/7 which makes the forget calls explicit, but keeps the check-closed-state logic with an assert.  Best of both worlds, maybe.

PR #223, opened 2020-02-03
DHCP timeout and tests

henry54809:


daq/host.py
```diff
@@ -257,6 +257,29 @@ def _prepare(self):
...
+        self.test_host.terminate()
+        self.test_host = None
+        self._docker_callback(exception=self._TIMEOUT_EXCEPTION)
+
+    def check_module_timeout(self):
+        """Checks module run time for each event loop"""
+        timeout_sec = self._get_test_timeout(self.test_name)
+        if not timeout_sec or not self.test_start:
+            return
+        timeout = gcp.parse_timestamp(self.test_start) + \
```
grafnu: use intermediate variable rather than line continuation

testing/test_aux.out
```diff
@@ -100,11 +103,15 @@ port-02 module_config modules
...
+  "dhcp": {
+    "timeout_sec": 120
+  },
   "hold": {
     "enabled": true 
   },
   "macoui": {
-    "enabled": true
+    "enabled": true,
+    "timeout_sec": 1
```
grafnu: looks like the macoui tests takes about ~1.3 seconds (just looking at the logs) -- so might be some flakyness here if some system decided to be fast.  If it works it works, but just something to watch out for.

testing/test_many.sh
```diff
@@ -18,17 +19,24 @@ echo sec_port=$((NUM_DEVICES+1)) >> local/system.conf
 ifaces=
 for iface in $(seq 1 $NUM_DEVICES); do
     ifaces=${ifaces},faux-$iface
-    echo autostart cmd/faux $iface >> $manystartup
+    xdhcp=""
+    if [[ iface -le $NO_DHCP_DEVICES ]]
+    then
```
grafnu: keep then on previous line (see use in test_aux.sh for example)

testing/test_many.sh
```diff
@@ -3,7 +3,8 @@
 source testing/test_preamble.sh
 
 NUM_DEVICES=8
-
+RUN_LIMIT=10
```
grafnu: With this change, can the run limit be increased?   I'm basically wondering if this change (dhcp timeout) has a measurable impact on the overall time to run the ~10 runs.

testing/test_many.sh
```diff
@@ -3,7 +3,8 @@
...
+NO_DHCP_DEVICES=1
```
grafnu: NUM_NO_DHCP_DEVICES -- otherwise I'd interpret this variable as a boolean, but you're using it in a comparison.

testing/test_many.sh
grafnu: you missed this one :-)

henry54809:
PATL

PR #227, opened 2020-02-06
Added DHCP lease command line option

henry54809:


daq/gateway.py
```diff
@@ -61,6 +61,7 @@ def _initialize(self):
         host = self.runner.add_host(host_name, port=host_port, cls=cls, tmpdir=self.tmpdir)
         host.activate()
         self.host = host
+        self.host.cmd('./change_leasetime %s' % self.runner.config.get("initial_dhcp_lease_time"))
```
grafnu: i'm torn if this should have units specificed, e.g., initial_dhcp_lease_sec so the value would be 500 vs. 500s.  Something about having "500s" seems... non-standard (I think it makes people warm and fuzzy to have time configuration parameters be integers). whatever, just something to consider. Not gonna stop anything either way.

But, alternatively, pick either _lease_time or _leasetime and be consistent :-)

misc/change_leasetime
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash -e
+#
+# Dynamically increase the DHCP lease time, requires killing and restarting
+# dnsmasq as per documentation (SIGHUP does not reload configuration file).
+lease=$1
+if [ -z $lease ]; then
+    lease="500s"
```
grafnu: I would make this error out -- no reason to have a default here and it can only cause confusion.

daq/gateway.py
henry54809: I didn't force the lease time to be seconds so any unit that dnsmasq supports can be passed in. It's also convenient to change to seconds if necessary in the future.

henry54809:
PTAL. Pending Travis build

misc/change_lease_time
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash -e
+#
+# Dynamically increase the DHCP lease time, requires killing and restarting
+# dnsmasq as per documentation (SIGHUP does not reload configuration file).
+lease=$1
+if [ -z $lease ]; then
```
grafnu: -z "$lease" -- otherwise it will be a syntax error if it's empty

misc/system.conf
```diff
@@ -89,3 +89,7 @@ site_path=local/site/
 
 # If defined, override destination root for device reports (defaules to {site_path})
 # site_reports=local/tmp/
+
+# DHCP lease time
+initial_dhcp_lease_time=120s
+dhcp_lease_time=500s
```
grafnu: consider "ongoing_dhcp_lease_time"

misc/change_lease_time
henry54809: Tested cases where $lease is an empty string or null; both cases worked fine without explicit string interpolation.

PR #228, opened 2020-02-07
Linux service 

henry54809:


bin/setup_daq
```diff
@@ -12,5 +12,10 @@ bin/setup_dev
 echo bin/build_daq
 bin/build_daq
 
+if [ -z $TEST ]; then
```
grafnu: I wouldn't make this go by default for now -- just leave it out of setup_daq -- not something that needs to happen all that often, and I think it will cause more problems then it would solve!

PR #229, opened 2020-02-11
Feature/stackdriver integration

henry54809:


bin/install_daq
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash -e
+
+ROOT=$(dirname $(realpath $0))/..
+sudo cp misc/daq.service /etc/systemd/system/
+sudo sed -i "s#%ROOT%#$ROOT#" /etc/systemd/system/daq.service 
+
+sudo systemctl enable daq.service
+sudo systemctl start daq.service
```
grafnu: should have trailing newline

misc/daq.service
```diff
@@ -0,0 +1,14 @@
...
+
+[Service]
+Type=simple
+Restart=always
+RestartSec=5
+User=root
+ExecStart=%ROOT%/cmd/run
+
+[Install]
+WantedBy=multi-user.target
```
grafnu: newline at EOF

.travis.yml
```diff
@@ -19,6 +19,7 @@ cache: pip
 env:
   global:
     - DOCKER_STARTUP_TIMEOUT_MS=60000
+    - TEST=1
```
grafnu: ?

daq/gcp.py
```diff
@@ -56,8 +57,16 @@ def __init__(self, config, callback_handler):
         self._storage = storage.Client(project=self._project, credentials=self._credentials)
         self._ensure_report_bucket()
         self._config_callbacks = {}
+        self._logging = logging.Client(credentials=self._credentials, project=self._project)
+
         LOGGER.info('Connection initialized at %s', get_timestamp())
 
+    def get_logging_client(self):
+        """Gets the stackdriver client"""
+        if hasattr(self, "_client_name"):
```
grafnu: what is this check for?  When is _client_name not defined?  If it's in the fail-out case when gcp_cred is not defined, then you should just set it to None in the constructor, so it will at least exist as an attribute (makes the check cleaner).  using hasattr is somewhat of a red-flag use...

henry54809:
PTAL. Pending travis build.

PR #231, opened 2020-02-13
 Wait for physical network interface to exist instead of exiting during startup

henry54809:


bin/physical_sec
```diff
@@ -58,7 +58,9 @@ if ! ifconfig $ext_ctrl || ! ifconfig $ext_intf; then
     echo Interfaces $ext_ctrl and $ext_intf not found, current set:
     ifconfig -a | fgrep flags
     echo
-    false
+    while ! ifconfig $ext_ctrl > /dev/null || ! ifconfig $ext_intf > /dev/null; do
+        sleep 20
```
grafnu: include echo message so people know what it's waiting for.

grafnu: also -- can you just replace the encompassing if?  I don't think they're both really needed...

henry54809: but there are 2 interfaces. Don't you have to assure both interfaces are there?
And ifconfig stderr prints out a message saying inquiring network interface is not found so I think another echo may not be necessary.

grafnu: Yes, but what does that have to do with there being both an if and a while?

if condition:
   while condition:
      foobar

the if is redundant

Good point about the echo.

henry54809: I misunderstood what you meant by encompassing if. The if statement doesn’t have /dev/null redirection, but I put that on the while statements to declutter the output a bit.

grafnu: I would go for elegance/correctness rather than just maintaining the status quo. Less important what used to be there, rather than what the desired behavior is.   Logically, I don't think the initial gating functionality (if statement) should be different than the retry functionality (while loop).  Whatever status/output is in one should be the same as the other...

henry54809:
PTAL. pending travis build.

PR #234, opened 2020-02-20
x230 ZTP

henry54809:


bin/ssh_switch
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/expect -f
+spawn ssh manager@192.168.1.1
```
grafnu: filename doesn't seem right.... I would expect ssh_switch to open a ssh connection to the switch, rather it's resetting factory default... suggest a better name!

PR #241, opened 2020-02-28
Feature/dockerhub

henry54809:


bin/docker_push
```diff
@@ -1,43 +1,25 @@
 #!/bin/bash -e
 
-# Travis CI seems to source this script, so it is ignoring the #! above!
```
grafnu: Why remove?

bin/docker_push
```diff
@@ -1,43 +1,25 @@
...
-    version_hash=$(git rev-list -n 1 $RTAG)
-
-    if [ $head_hash != $version_hash ]; then
-	echo Release head does not match version tag.
-	echo head=$head_hash
-	echo rtag=$version_hash
-	false
-    fi
-		   
+if [ -n "$TARGET" ] && [ -n "$VERSION" ]; then
```
grafnu: TARGET and VERSION seem a bit too generic for this behavior -- they could very easily overlap with something else and cause problems -- prefix with DAQ_ to provide a namespace for them...

bin/test_daq
```diff
@@ -31,13 +31,13 @@ if [ -d faucet ]; then
 fi
 
 if [ -d inst ]; then
-    echo Directory inst/ exists, clean out before test run.
-    false
+    echo Directory inst/ exists, cleaning out before test run.
+    mv inst "inst.$(date)"
```
grafnu: why not the hard fail?  It was meant to a bad system test state.  When does this happen that you want to just move it?

cmd/build
```diff
@@ -1,51 +1,48 @@
 #!/bin/bash -e
+# The following are expected modes of builds
+# cmd/build <- Builds all docker images
```
grafnu: hmmm, I would make sure these are coherent.... right now, you have:
push == push
pull == pull then build
_nothing_ == build all

I would make it more explicit, so:
pull == pull
build == build all
_nothing_ == pull then build

testing/test_aux.sh
```diff
@@ -160,7 +160,7 @@ echo Redacted docs diff | tee -a $TEST_RESULTS
     | tee -a $TEST_RESULTS
 
 # Make sure there's no file pollution from the test run.
-git status --porcelain | tee -a $TEST_RESULTS
+git status --porcelain -u no | tee -a $TEST_RESULTS
```
grafnu: The entire purpose of this was to make sure there was no pollution side-effects... what is it specifically that you're trying to ignore?

.gitignore
```diff
@@ -134,3 +134,4 @@ ENV/
 
 # other
 testing/*dream*
+misc/docker_images.txt
```
grafnu: why ignore this?

bin/docker_push
henry54809: I'm not seeing this problem so I removed it to avoid confusions.

cmd/build
henry54809: I'll change the behavior to follow this.

testing/test_aux.sh
henry54809: It's always writing to misc/docker_images.txt which I ignored explicitly in .gitignore. I also don't want this file to be changed accidentally.

bin/test_daq
henry54809: the config_base.sh writes out to inst/ before the test is run. I don't see the necessity to have one more environmental variable / other logic to indicate that config_base.sh should not write out to inst. I think it's more straight forward to backup users' old inst and local folders and create new ones to keep the test running.

bin/docker_push
grafnu: But now -e is specified in two places with not reasoning... so maybe take out the "set -e" line as well.

bin/test_daq
grafnu: What does this have to do with the docker images?

testing/test_aux.sh
grafnu: Ok -- so -- my question is likely then why are you ignoring that file?   How do you know which images to download?

bin/test_daq
henry54809: Currently there are 2 places where cmd/build is called. Once during setup time, once during aux test time. Since the correct docker images to be built are determined by the local system config, there needs to be a local and inst folders for that to work. This means, after running setup, there will be inst and local folder before cmd/run is called. Previously, you have specific logic to determine if the cmd/build is called under setup and selectively disable creating the inst and local folders so these checks from above will pass. I think that extra logic makes the workflow more complicated to understand.

testing/test_aux.sh
henry54809: I have committed the file then ignored it. The file is technically untracked under this condition so it won't show up under git status -u no.

grafnu:
Ooohhh ug. What's the fundamental reason for that? Doing something like
that (committed then ignored) can be really confusing because it messes
with trying to diagnose or understand anything. Sometime maybe we can have
a think about the flow here... I'm not quite sure I understand what's going
on... seems to me there should be a clean way to do all this, but I might
be missing something.

grafnu:
I think there might be something fundamental about the flow that can be
fixed to make it cleaner rather than some of this.  Seems to me like it
only needs to be called once, or calling it multiple times should be
idempotent unless configs change. Maybe the assertion around having the
docker build called when building daq is wrong... or something like that.
Like the other point about the ignored file -- these changes smell like
there's something unclean about the basic model, so would like to
understand and fix that rather than trying to patch it over.

henry54809:
PTAL

cmd/build
```diff
@@ -1,51 +1,63 @@
...
-    do_images=load
+if [ "$1" == push ]; then
+    cat $DOCKER_IMAGE_LIST | while read image hash other; do
+        found=$(docker images $image | awk '{print $3}' | grep $hash) || echo ""
+        if [ -z "$found" ]; then
+            echo $image:$hash not found locally...
+            continue
+        fi
+        echo Push $image:$hash to dockerhub...
+        TARGET="$image" VERSION="$hash" ./bin/docker_push
```
grafnu: Why aren't these command line arguments?  This seems very odd.  Where else is docker_push used?

cmd/build
```diff
@@ -63,114 +75,14 @@ if [ "$target_commit" != "$local_version" ]; then
...
-        true
-    }
-    function remove {
-        true
-    }
-
-    include ${host_tests:-misc/host_tests.conf}
-)
-docker_files=$(<$build_list_file)
+CHECK_EXISTING=$CHECK_EXISTING bin/docker_build $inline
```
grafnu: I would convert this to a command-line arg, unless there's a reason to be env arg.

bin/docker_push
grafnu: Is this called direct by travis?  If not, it should be on the first line like all the other files (just convention)

cmd/build
henry54809: I just think this is a lot cleaner and easier to use an env var right here.

grafnu:
Why?  It's completely nonstandard. How is

CHECK_EXISTING=$CHECK_EXISTING bin/docker_build $inline

cleaner than

bin/docker_build $CHECK_EXISTING $inline

?

henry54809:
@grafnu  for bin/docker_build $CHECK_EXISTING $inline
you have to check the existence of CHECK_EXISTING and inline, where there are a total of 4 possible cases.

grafnu:
No you don't. They can be treated independently.
https://github.com/grafnu/forch/blob/master/bin/setup_stack#L9

henry54809:
PTAL

PR #252, opened 2020-03-09
Cleanup network module

grafnu:


subset/network/network_tests.py
```diff
@@ -153,22 +154,22 @@ def test_communication_type_broadcast():
...
+        add_summary("Unicast packets received. ")
     return 'info'
 
 def test_ntp_support():
     shell_result = shell_command_with_result(tcpdump_display_ntp_packets, 0, False)
     ntp_packets_received = packets_received_count(shell_result)
     if ntp_packets_received > 0:
-        add_summary("NTP packets received.\n")
-        add_packet_info_to_report(ntp_packets_received, packet_request_list)
+        add_summary("NTP packets received. ")
```
henry54809: isn't it easier to include the empty space after every add_summary call in the add_summary function instead? 
so on line 172: 
summary_text += text + " "

grafnu: Done

PR #255, opened 2020-03-11
adding options doc and change build options to use pull

henry54809:
docker_images file is also updated to match changes in a previous PR.


cmd/run
```diff
@@ -16,7 +16,7 @@ sudo rm -rf inst/reports inst/run-port-* $cmdrun_log
 sudo chown -f $USER -R inst || true
 
 if [ -n "$build_tests" ]; then
-    cmd/build
+    cmd/build check
```
grafnu: actually, now that I look at it, this logic is wrong.  By definition, -b should run cmd/build -- it should actually build the tests.   The pulling images from docker behavior is broken right now, so can't be enabled by default until it's properly planned out with a complete build flow...

PR #257, opened 2020-03-11
Intermediate fixes for build setup

grafnu:


henry54809:
So integration tests on travis will continue to build all docker images?

henry54809:
how about adding the "-b" options doc here in the PR as well so I can delete my other PR?

`-b`, `build_tests`: Pull docker images for tests, then build the ones that are missing. Runs "cmd/build"

grafnu:
Updated with documentation change.

Yes, the theory is that for now it builds always, which keeps it logically consistent, although can be slow. But, the ability to push/pull images is there there, so it can at least be used for release versions, e.g. DRA or Qualification. But, that will still take some work to formalize.  Then also, while we're sorting out the entire dev pipeline we can see how docker images fit in with that -- but I don't think that's ready to be the default behavior just yet.

PR #260, opened 2020-03-11
Fix container termination exception

grafnu:


grafnu:
Not ready yet--- still failing but just in a different way.

PR #261, opened 2020-03-12
security.password

em-redstone:
@grafnu PTAL

Hi Trevor,

Good afternoon, this pull request contains the latest changes for the password test.

The changelog mostly involves:
- Added test coverage in all pass/fail/skip cases
- SSH/Telnet now working, and now use hydra
- Protocol tests now run async and write to separate log files
- Test log added to device report
- Code cleanup and refactor

Many thanks!

Edzel.


misc/start_faux
```diff
@@ -58,6 +57,9 @@ if [ -n "${options[passwordfail]}" ]; then
     /usr/sbin/sshd -D &
     service xinetd start
     service ssh start
+else
```
grafnu: is stop needed?  Or, if stop is needed, is start needed?

misc/start_faux
```diff
@@ -160,6 +163,8 @@ elif [ -n "${options[expiredtls]}" ]; then
     echo Starting expired tls server on port 443 https.
     mv /root/nginx/nginxfail.conf /etc/nginx/nginx.conf
     service nginx start
+else
+    service nginx stop
```
grafnu: ditto

reports/ssh_report.txt
```diff
@@ -0,0 +1 @@
+RESULT fail security.passwords.ssh Default passwords have not been changed.
```
grafnu: ??

subset/security/security_passwords/src/main/test_docker_files/Dockerfile.faux_fail
```diff
@@ -0,0 +1,31 @@
+# Dockerfile for running the corresponding faux device.
```
grafnu: What's this file for?  This doesn't seem right...

misc/start_faux
em-redstone: Hi Trevor, these were clauses I felt were needed before the refactor to make the code logic work, but after running the tests again with the old start_faux file, I've found I no longer need these. Will remove these on next commit.

misc/start_faux
em-redstone: ditto

reports/ssh_report.txt
em-redstone: This was made by accident when testing the code, thanks for spotting this. Will also remove this!

subset/security/security_passwords/src/main/test_docker_files/Dockerfile.faux_fail
em-redstone: The files in test_docker_files/ folder were local tests I made to understand the code logic, and also to test my changes a bit quicker. They actually serve no purpose other than testing the code and can be removed with no effect. I just left it in as it might help others working on this test in the near future. 

Thanks!

grafnu:
Ok - - yes, we have plans to make a utility for better testing things more
quickly in the future for any test.  But, for now likely best to remove
unless there's significant additional unique code (which I don't think
there is after a quick glance).

em-redstone:
@grafnu PTAL

Hi Trevor,

The changes recommended have been applied.

Kind regards,

Edzel.

PR #262, opened 2020-03-12
Docs update

henry54809:
Please review my updates to docs.


docs/quickstart.md
```diff
@@ -1,24 +1,55 @@
 To get up and going with the DAQ basics, all you need to do is...
 
+## Install
 (starting with a vanilla Debian Linux install)
 <pre>
 # git clone http://github.com/faucetsdn/daq.git
 &hellip;
-# cd daq
-# bin/setup_daq
+cd daq
```
grafnu: The # at the front was the distinguisher from "command prompt" -- if not #, I think it should be something... otherwise you can't tell the difference from command output.

docs/quickstart.md
```diff
@@ -1,24 +1,55 @@
...
+
+## Test Run
+<pre>
+cmd/run -s
 &hellip;
 Done with run, exit 0
 </pre>
+Reports generated fall under **inst/reports**. An example report can be viewed [here.](report.md)
+
+***For more run options please reference [here.](options.md)***
```
grafnu: nit, prefer to just link the words, not a "here" -- e.g. 

***The system can be further configured by using a [variety of run options](optiond.md)***

docs/quickstart.md
henry54809: i think I'll use a $ instead. # is more like a comment to me.

grafnu:
yes -- I usually use $ too, not sure why this time around it ended up #

henry54809:
PTAL

PR #305, opened 2020-03-19
Use docker hub image flow for tagged releases

grafnu:


grafnu:
> Can the docket_images.ver file be an ephemeral one rather than a permanent file stored inside misc?

It should be permanent file, but... it shouldn't disrupt the flow. It's only used when working with a specifically tagged release version, otherwise it's ignored and images are built. The model is that "if you're working with a release version, it pulls images from dockerhub, but as soon as you start editing anything then it reverts to build mode" -- that file is just use for sanity checking to make sure your tags are up to date.

Is there a particular scenario you're concerned about?

pisuke:
Thanks for clarifying the working model.

> Is there a particular scenario you're concerned about?

As you mentioned it does feel a bit like it's disrupting the workflow, as it pops up as a file to be committed because it gets recreated after building the docker images.
But maybe it's not that big of a deal after all. If it doesn't bother you, all good.

grafnu:
The flow is more subtle than that.  WHen you build, it gets put in inst/ so
is not picked up by git. Only when you push images to github does it update
the files in misc/ -- so, it's separating out a development flow from a
release flow.

pisuke:
Nice, now makes perfect sense.
I'll try it again, sounds like the behaviour has been changed in the last few days. :-)

PR #310, opened 2020-03-20
Use docker images for tagged releases

grafnu:


grafnu:
Ok -- fixed. All build paths seem to be working.

PTAL

Issue #247, opened 2020-03-05
Develop the security.tls.v1 test (test#47)


pisuke:
The following tests must be carried for validation.
1. Minimum version is V1 (this means TLS 1.0) - ERROR if not
2. RSA public key length is 2048 bit minimum - ERROR if not
3. Check if the cipher suite name includes both ECDH and ECDSA with minimum bit length of 224 - WARNING if not.

pisuke:
@jhughesbiot can you please pick up developing this test? I've put a starting point for the test development and validation. Thanks!

jhughesbiot:
There seems to be some recent updates that have severely caused issues with working with this test.  test_aux.sh runs against 3 aux devices, only two generate reports and there is no output of the security module in any of the run nodes to compare what the actual vs expected results are.  Whatever update broke this ability will likely need to be corrected before I can continue troubleshooting this issue.

jhughesbiot:
Pulled the most recent code updates and built and reran test_aux.sh.  None of the 3 aux devices seem to be configured to even run the security module anymore so they won't test existing tls tests and aux device one 9a02571e8f01 always completely fails and stops all tests after its ping failure.

INFO:runner:Target port 1 finalize: ['01:ping:Exception'] (Test initialization failure)
INFO:runner:Target port 1 cancel 9a:02:57:1e:8f:01 (#3/0).

There needs to be a rollback somewhere that broke the CI.

grafnu:
Can you point at the specific spot in a Travis test run where you're seeing
the bad behavior?  I looked at the faucetsdn/daq master runs
<https://travis-ci.com/github/faucetsdn/daq/branches> and they seem ok
(except for the flaky security.passwords.telnet test itself).   I think you
may be misinterpreting some of the log output, but I can't tell for sure
because I don't know exactly what you're looking out. Yes, there is a
"ping:Exception" error in there somewhere, but that's testing the ping
exception specifically -- and there's another test run that tests other
things.  From what I can tell things are working fine, so I'd need more
specific information about exactly what you think is broken (e.g. point me
at a Travis run that has the logs).

If Travis is behaving and it's a local problem then that's something else
entirely, but that doesn't seem to be what you're describing.

jhughesbiot:
[...]

Issue #213, opened 2019-11-05
TLS tests report V3 success without testing version


jhughesbiot:
The TLS tests in the security module report for V3 support  but do not tests v3 only for valid certificates.

pisuke:
Addressed in #551, #552, #626

PR #317, opened 2020-03-23
Update TLS methods and tests

jhughesbiot:
Add TLS 1.0 support
Add RSA public key bit length validation
Add cipher key checks, warn only

Initial PR against grafnu for the TLS 1.3 additions: https://github.com/grafnu/daq/pull/269

Addresses Issues: https://github.com/faucetsdn/daq/issues/247 and https://github.com/faucetsdn/daq/issues/213


subset/security/Dockerfile.test_tls
```diff
@@ -1,9 +1,9 @@
-FROM daqf/aardvark:latest
+FROM daq/aardvark:latest
```
grafnu: this should be daqf/ -- something we've changed recently. DockerHub requires 4-letters minimum for a project name (and we're working on a distribution framework that would use pre-built docker images).  If daqf/ doesn't work for you for some reason, please let me know and we can fix.

jhughesbiot:
@grafnu @pisuke 
Can you take a look at this and provide some feedback?

subset/security/module_manifest.json
```diff
@@ -1,6 +1,12 @@
 {
-    "security.tls.v3" : {
-        "desc" : "Verify the device supports TLS 1.2 or above (as a client)"
+    "security.tls.v1_3" : {
+        "desc" : "Verify the device supports TLS 1.3 (as a client)"
```
grafnu, line 3: should be "description" -- no need to abbreviate here -- I don't think we have any schema enforcement for this yet...

testing/test_aux.out
```diff
@@ -12,12 +12,26 @@ RESULT info protocol.bacnet.version Protocol version: 1
...
+RESULT fail security.tls.v1_3.x509
+RESULT pass security.tls.v1
+RESULT pass security.tls.v1.x509
+RESULT pass security.tls.v1_2
+RESULT pass security.tls.v1_2.x509
+RESULT fail security.tls.v1_3
+RESULT fail security.tls.v1_3.x509
+RESULT pass security.tls.v1_2
+RESULT pass security.tls.v1_2.x509
+RESULT fail security.tls.v1_3
```
grafnu, line 33: no skip result (what about not being configured?) and also v1_3 fails all the time?

testing/test_modules.out
```diff
@@ -5,12 +5,19 @@ RESULT skip base.startup.dhcp Using static ip 10.20.0.5
 RESULT skip base.switch.ping No local IP has been set, check ext_loip in system.conf
 RESULT pass base.target.ping target reached %% 10.20.0.5
 Testing tls
-RESULT skip security.tls.v3
-RESULT skip security.x509
+Module execution failed.
```
grafnu, line 8: why is the module failing?  That doesn't seem right!

testing/test_aux.out
jhughesbiot: Are any of the faux devices configured to support 1.3?  This is just the test result of the existing device configurations so if they weren't written to support it, it will always fail.

testing/test_modules.out
jhughesbiot: What determines a module to pass or fail?  This seems to happen outside of the tls test so I can't really comment on what is determining this.

subset/security/module_manifest.json
jhughesbiot: Updated

grafnu:
Then why is it changing?  If it's outside of the tls test then I would
expect no difference.   Are the security tests something you were working
with or changing?  It seems like they came from the tls module, no?

If you just run "bin/test_module tls" what does it output?

Module failure is determined by a non-zero exit status from the container,
which should only happen if there's some procedural error (not test
failure).

grafnu:
I don't know, but that's part of putting the module together... most/all of
these need to have pass/fail/skip results b/c those are the three main
states. Otherwise, we don't really know if the test is functioning properly
or not. Before, there was:

RESULT pass security.tls.v3

Is this something different?

jhughesbiot:
Yes, tls.v3 is not the same as TLS 1.3, it supports up to 1.2 but you don't get to control what version is being tested, just that one of the supported TLS versions with that configuration was accepted.  That's what this update does, breaks out each version and tests independently.  The tls.v3 was also using java 8 which doesn't support TLS 1.3 either, which this also updates.

jhughesbiot:
After inspecting the messaging from bin/test_module tls, I have found that the module fails. Running bin/test_module tls tls, does not show the same module error.  There is no uncaught errors being thrown, only caught exceptions that are printed to the output for logging.

It appears this is being handled by this piece of code  from bin/test_module but i'm not sure what this is all doing to determine a pass/fail of a module: 

```
run_cmd=$(docker inspect "--format={{json .Config}}" $image | jq -r '(.Cmd | join(" "))')
echo
echo
echo %%%%%%%%%%%%%% Running module command $run_cmd %%%%%%%%%%%%%%%
failed=
if [ -z "$cmd_exec" ]; then
    docker exec $container $run_cmd || failed=yes
else
    docker exec -ti $container bash || failed=yes
fi

if [ -n "$failed" ]; then
    echo
    echo Module execution failed.
    false
fi
```

grafnu:
If you try "bin/test_module -e tls", then it should drop into an execution
shell. From there, you should be able to run "bash -ex ./test_tls" which
will show what the command is doing and when it fails -- see below for what
I see when I run it.  I just noticed this line in there:

++ jq --arg tn security.x509 -r '.[$tn].desc' ./module_manifest.json

which might be failing now because I told you to use "description" --
so maybe try changing that back to see if it fixes that problem.  If
that's the case, then I'll have to work with Francesco to rectify
that.

    docker exec $container $run_cmd || failed=yes

That bit of code is running the module start command, then remembering if
it fails. No magic there, it's simply capturing the result of the run and
propagating the error state.

%%%%%%%%%%%%%% Running module command ./test_tls %%%%%%%%%%%%%%%
root@daq-tls:~# bash -ex ./test_tls
+ source reporting.sh
++ _______REPORT_DIVIDER=--------------------
+ REPORT=/tmp/report.txt
+ LOCAL_REPORT=tmp/report.txt
+ RESULT_LINES=/tmp/result_lines.txt
+ MONO_LOG=/tmp/mono-log.txt
+ MANIFEST=./module_manifest.json
+ '[' -n 10.20.0.5 ']'
+ echo Collecting TLS cert from target address %% 10.20.0.5
+ java -jar tlstest/build/libs/tlstest.jar 10.20.0.5
getCertificate IOException:Connection refused (Connection refused)
report:IOException unable to connect to server.
RESULT skip security.tls.v3
RESULT skip security.x509

Certificate read failed
+ grep RESULT tmp/report.txt
+ grep -v RESULT tmp/report.txt
+ write_out_monolog /tmp/report.txt ./module_manifest.json
/tmp/mono-log.txt /tmp/result_lines.txt
+ local _REPORT=/tmp/report.txt
+ local _MANIFEST=./module_manifest.json
+ local _MONO_LOG=/tmp/mono-log.txt
+ local _RESULT_LINES=/tmp/result_lines.txt
+ mapfile -t _TEST_ARR
++ jq -r 'keys[]' ./module_manifest.json
+ echo --------------------
+ tee -a /tmp/report.txt
--------------------
+ cat /tmp/mono-log.txt
+ tee -a /tmp/report.txt
Collecting TLS cert from target address %% 10.20.0.5
IOException unable to connect to server.
+ tee -a /tmp/report.txt
+ echo

+ for test_name in "${_TEST_ARR[@]}"
++ jq --arg tn security.tls.v3 -r '.[$tn].desc' ./module_manifest.json
+ test_desc='Verify the device supports TLS 1.2 or above (as a client)'
++ grep 'RESULT.*security.tls.v3' /tmp/result_lines.txt
+ write_out_result /tmp/report.txt security.tls.v3 'Verify the device
supports TLS 1.2 or above (as a client)' 'See log above' 'RESULT skip
security.tls.v3'
+ local REPORT=/tmp/report.txt
+ local TEST_NAME=security.tls.v3
+ local 'TEST_DESCRIPTION=Verify the device supports TLS 1.2 or above (as a
client)'
+ local 'LOG=See log above'
+ local 'RESULT_AND_SUMMARY=RESULT skip security.tls.v3'
+ cat
+ for test_name in "${_TEST_ARR[@]}"
++ jq --arg tn security.x509 -r '.[$tn].desc' ./module_manifest.json
+ test_desc='Verify the devices supports RFC 2459 - Internet X.509 Public
Key Infrastructure Certificate and CRL Profile'
++ grep 'RESULT.*security.x509' /tmp/result_lines.txt
+ write_out_result /tmp/report.txt security.x509 'Verify the devices
supports RFC 2459 - Internet X.509 Public Key Infrastructure Certificate
and CRL Profile' 'See log above' 'RESULT skip security.x509'
+ local REPORT=/tmp/report.txt
+ local TEST_NAME=security.x509
+ local 'TEST_DESCRIPTION=Verify the devices supports RFC 2459 - Internet
X.509 Public Key Infrastructure Certificate and CRL Profile'
+ local 'LOG=See log above'
+ local 'RESULT_AND_SUMMARY=RESULT skip security.x509'
+ cat

grafnu:
https://github.com/faucetsdn/daq/pull/346

jhughesbiot:
@grafnu 

As expected, the module failure is something other than what to do with the errors within the test. When I run from within the shell, the module does not fail, there is no error to be found:

> RESULT skip security.tls.v1
> RESULT fail security.tls.v1.x509
> RESULT skip security.tls.v1_2
> RESULT fail security.tls.v1_2.x509
> RESULT skip security.tls.v1_3
> RESULT fail security.tls.v1_3.x509

> Certificate read failed

When running just the module test, which includes the module test code that exists outside of the actual module:

> bin/test_module tls:


> Certificate read failed
> 
> **Module execution failed.**

I am also experiencing a new merge conflict I cannot seem to get past since the last commit.  Either way, this issue is kind of stuck right now with whatever code is failing the module.

grafnu:
I think the problem is one of the grep lines that are erroring out. If you
run it with shell tracing on, you can see where/why it stops (this is
running as your branch):

$ *bin/test_module -e tls*

...

%%%%%%%%%%%%%% Running module command ./test_tls %%%%%%%%%%%%%%%
root@daq-tls:~#* bash -ex ./test_tls *
+ source reporting.sh

...

RESULT skip security.tls.v1_3
RESULT fail security.tls.v1_3.x509

Certificate read failed
+ grep RESULT tmp/report.txt
+ grep -v RESULT tmp/report.txt
root@daq-tls:~#

So, I don't know exactly what those lines are for, but that grep statement
is why it's erroring out and claiming module failure.

jhughesbiot:
I am at a loss on how to proceed at this point.  I resolved all conflicts but I need to update the module manifest so this can work to the new format..  Once I attempt to upload the changes again, it's back to a conflict and so the CI won't run.

grafnu:
I'm afraid you're going to have to be more specific about the conflicts.
Your merge-base is two weeks old at this point so that might be why you're
having conflicts.

~/daq$ git log -n 1 `git merge-base 1.1.0 jhughes/tls-updates`
commit 474cadae30081627256d83bcd7dd8020404f2a61
Author: Trevor <grafnu@users.noreply.github.com>
Date:   Wed Mar 25 11:06:02 2020 -0700

    Pre-release cleanup (#322)

jhughesbiot:
I continually keep pulling from master just to keep things up to date and new conflicts keep popping up as new things are merged into master.  Before it was just the module manifest, now there's 2 more files.  I really want to get this PR complete but I think I might just have to close it and re-open it with the changes I care about only as the current merging deployment is causing this  to be an unmanageable process.

pisuke:
Hi @jhughesbiot I share your pain and I was in the same place as you with the nmap PR a couple of weeks ago. This [article on rebasing](https://www.atlassian.com/git/tutorials/git-forks-and-upstreams) helped me figure out a better way to sync to upstream using rebase. When there is little happening on the master branch doing the pull and merge is not a problem, but there have been so many things happening recently on master that rebasing seems to be the best option. Hope it's useful!

pisuke:
Addressed in #551, #552 and #626

PR #322, opened 2020-03-25
Pre-release cleanup

grafnu:


cmd/build
```diff
@@ -86,6 +90,11 @@ elif [ "$1" == force ]; then
     fi
     force=force
     shift
+elif [ "$1" == "" ]; then
+    echo Defaulting to force build
+else
+    echo Unknown command $1
```
henry54809: This will break ability to do cmd/build inline. since inline will be $1 at this point.

grafnu:
Updated, PTAL

PR #324, opened 2020-03-25
Add password switch config

jhughesbiot:


jhughesbiot:
@grafnu @pisuke 
Can you take a look at this and provide some feedback?

jhughesbiot:
@grafnu @pisuke 
Any thoughts on this one?

daq/docker_test.py
```diff
@@ -48,6 +48,9 @@ def start(self, port, params, callback):
                          "SWITCH_PORT=" + params['switch_port'],
                          "SWITCH_IP=" + params['switch_ip'],
                          "SWITCH_MODEL=" + params['switch_model']]
+        if 'switch_username' in params:   
+            env_vars += ["SWITCH_USERNAME=" + params['switch_username'],
```
pisuke, line 52: This is a nice feature. 
I was thinking originally that it could be put inside one of the `module_config.json` when configuring the switch module, something like

```
    "switch": {
      "enabled": true,
      "switch_model": "CISCO_9300",
      "username": "admin",
      "password": "password",
      "poe": {
        "enabled" : true
      }
    },
```

rather than in the `system.conf` file. @grafnu @jhughesbiot what do you think?
If you both think using `system.conf` is better, then we should move the poe boolean inside `system.conf` too. 
It would be best to have all the configuration variables for the test in the same place.

grafnu:
Francesco, I think you are right... I think that thought crossed my mind at
one point but then I forgot about it. It should be easy enough to fix the
code, I can do that... but it will break a few of the existing setups but I
think we'll just have to accept that hit.

The only change I might consider is to uplevel the information so it's not
inherent to the switch *module* -- e.g., the path right now is
modules/switch/poe/enabled: true -- but there's a world where there might
be another modules that wants the same information, so the path would just
be switch/poe/enabled: true -- not a big difference, but semantically keeps
things isolated better!

pisuke:
@jhughesbiot are you able to implement the suggested change?

grafnu:
Francesco, let me do this one... I think it'll reach into how some things
are configured and I wouldn't want to distract Jonathan (further) from the
tests themselves.

jhughesbiot:
When the suggestion to move the update into the module_config.json, was this meant at the device level?  This means a lot of duplicate effort to configure every device tested with a switch to have the same switch information.  With PoE, this makes sense as it is very specific to a device as some will and won't have it.  Am I wrong in interpreting the location of the config to be used?  Or is this more of a proposal to create a new switch_config.json file elsewhere that can be used globally?

grafnu:
The config mechanism is layered... so the config presented to the test
module is a composite of other config sources -- some are global, install,
etc... Either way, this is about the test *module* correct? And that
configuration will be independent from the device.  The site_path docs
describe some of this, but not everything (since it hasn't all been
implemented yet):
https://github.com/grafnu/daq/blob/master/docs/site_path.md

PR #330, opened 2020-03-30
glinux setup script

henry54809:


bin/setup_glinux
```diff
@@ -0,0 +1,43 @@
...
+  "data-root": "/usr/local/google/docker",
+  "bip": "192.168.9.1/24",
+  "storage-driver": "overlay2",
+  "debug": true,
+  "registry-mirrors": ["https://mirror.gcr.io"]
+}
+EOF
+
+sudo service docker start
+sudo addgroup docker || true
```
grafnu: this part seems generic, and not specific to gLinux -- should it be in the common part of the script?

henry54809: do you mean the addgroup docker part? These are also in the bin/setup_base.

grafnu: Yes -- so why is it in both?

henry54809: I just wanted that script to also be standalone.

grafnu: fair 'nuff

PR #343, opened 2020-04-01
Update host path

grafnu:
This adds a bit more structure to the data files stored.

### DIFF
--- a/daq/host.py
+++ b/daq/host.py
@@ -177,10 +177,9 @@ class ConnectedHost:
         return self._loaded_config['modules'].get('ipaddr', {}).get('dhcp_mode')
 
     def _get_unique_upload_path(self, file_name):
-        unique_name = os.path.basename(file_name)
-        if self.test_name:
-            unique_name = "%s-%s.%s" % (self.target_port, self.test_name, unique_name)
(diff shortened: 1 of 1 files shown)


daq/host.py
```diff
@@ -177,10 +177,9 @@ def _get_dhcp_mode(self):
         return self._loaded_config['modules'].get('ipaddr', {}).get('dhcp_mode')
 
     def _get_unique_upload_path(self, file_name):
-        unique_name = os.path.basename(file_name)
-        if self.test_name:
-            unique_name = "%s-%s.%s" % (self.target_port, self.test_name, unique_name)
-        return os.path.join(self.run_id, unique_name)
+        base = os.path.basename(file_name)
+        partial = os.path.join('tests', self.test_name, base) if self.test_name else base
+        return os.path.join('run_id', self.run_id, partial)
```
henry54809, line 182: what's the "run_id" folder for? are there going to be other folders under test site name folder?

grafnu:
Maybe, it's future-proofing the structure. Also, follows general path
naming conventions where if you have a collection of things then it should
be prefaced by a descriptive name. Since they are run_id's they should be
in run_ids/{run_id} regardless of what else might be there.

PR #345, opened 2020-04-02
dhcp cleanup and misc test improvements

grafnu:
Also makes the tests more DHCP-specific, rather than trying to do the same as test_many

### DIFF
--- a/testing/test_aux.sh
+++ b/testing/test_aux.sh
@@ -96,9 +96,6 @@ cmd/build inline
 echo Starting aux test run...
 cmd/run -s -k
 
-# Check custom timeout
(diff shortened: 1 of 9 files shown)


testing/test_aux.sh
```diff
@@ -96,9 +96,6 @@ cmd/build inline
 echo Starting aux test run...
 cmd/run -s -k
 
-# Check custom timeout
-cat inst/cmdrun.log | grep "Monitoring timeout for macoui after 1s" | tee -a $TEST_RESULTS
```
henry54809, line 100: I think removing this is gonna cause aux test to fail.

grafnu: Whelp. You were right... I could have sworn I checked that.

BTW, the functionality is still tested without this -- it shows up in several places when the macoui (and other tests) fail because of the timeout...

Issue #304, opened 2020-03-19
sed log errors in testing


grafnu:
as per [this travis run](https://travis-ci.com/github/henry54809/daq/jobs/299690753#L4324), there's some sed errors appearing in the logs:

    sed: -e expression #8, char 34: Invalid preceding regular expression
    sed: -e expression #8, char 34: Invalid preceding regular expression

looks like a bad sed expression creeped in and should be cleaned up.

pisuke:
OK, noted. Let me have a look.

OzyOzk:
Seems to be caused by the redacted function in test_aux.sh

```
function redact {
    sed -E -e "s/ \{1,\}$//" \
        -e 's/\s*%%.*//' \
        -e 's/[0-9]{4}-.*T.*Z/XXX/' \
        -e 's/[a-zA-Z]{3} [a-zA-Z]{3} [0-9]{1,2} [0-9]{1,2}:[0-9]{1,2}:[0-9]{1,2} [0-9]{4}/XXX/' \
        -e 's/[0-9]{4}-(0|1)[0-9]-(0|1|2|3)[0-9] [0-9]{2}:[0-9]{2}:[0-9]{2}\+00:00/XXX/' \
        -e 's/DAQ version.*//' \
        -e 's/[0-9].[0-9]{2} seconds/XXX/' \
        -e 's/\b(?:\d{1,3}\.){3}\d{1,3}\b/XXX/'
}
```
In the eighth expression there is a non-capturing group (?:\d{1,3}\.){3}\d{1,3} which are not supported by sed.

pisuke:
Thanks @OzyOzk, this regex is intended to replace IP addresses with the XXX string so that the output files can be matched.
Have you already thought of a regex replacement?
Maybe this one could work 
sed -e 's/\([0-9]\{1,3\}\.\)\{3\}[0-9]\{1,3\}/XXX/'

OzyOzk:
@pisuke I removed the ?: from statement 8 and tested on various regex testers. It removes the IP as before. I noticed that  before making this change the redacted device reports were all empty.

```
cat docs/device_report.md | redact > out/redacted_docs.md
cat inst/reports/report_9a02571e8f01_*.md | redact > out/redacted_file.md
```

On the above two lines both redacted files were empty as the redact function was failing. So when test_aux compares them to find the difference it was always reporting that there was no difference in the file, giving false positives for all aux tests carried out since this error crept in.

pisuke:
[...]

PR #349, opened 2020-04-03
#304 sed log errors

OzyOzk:
The redact function no longer throws a sed error. The current test_aux fail is caused by a change to the output of the nmap test, causing it to be different to the output in docs/device_report.md which was likely not detected due to the sed error.

```
Redacted docs diff
125c125
< # Nmap 7.60 scan initiated XXX as: nmap -v -n -Pn -T5 -sU -sT --open -pU:47808,T:23,443,80, -oG /tmp/nmap.log 10.20.73.164
---
> # Nmap 7.60 scan initiated XXX as: nmap -v -n -T5 -sT -sU --host-timeout=4m --open -pU:47808,T:23,443,80, -oG /tmp/nmap.log 10.20.3.164
127,128c127,128
< Host: 10.20.73.164 ()	Status: Up
< Host: 10.20.73.164 ()	Ports: 47808/open|filtered/udp//bacnet///	Ignored State: closed (3)
---
> Host: 10.20.3.164 ()	Status: Up
> Host: 10.20.3.164 ()	Ports: 47808/open|filtered/udp//bacnet///	Ignored State: closed (3)
```

https://travis-ci.org/github/OzyOzk/daq/jobs/670552789


testing/test_aux.sh
```diff
@@ -146,7 +146,7 @@ function redact {
         -e 's/[0-9]{4}-(0|1)[0-9]-(0|1|2|3)[0-9] [0-9]{2}:[0-9]{2}:[0-9]{2}\+00:00/XXX/' \
         -e 's/DAQ version.*//' \
         -e 's/[0-9].[0-9]{2} seconds/XXX/' \
-        -e 's/\b(?:\d{1,3}\.){3}\d{1,3}\b/XXX/'
+        -e 's/\b(\d{1,3}\.){3}\d{1,3}\b/XXX/'
```
henry54809, line 149: Actually this doesn't replace the ip in sed. Sed doesn't have the \d set, so you will have to expand the \d yourself into [0-9]. This is also why you're seeing the ip address diff in your test_aux run. Now it looked like test_aux wasn't running properly when the redact was invalid, so the diff you are seeing now was always there but it wasn't failing the aux test.  This should be fixed before merging in.

grafnu:
I fixed this as part of https://github.com/faucetsdn/daq/pull/351 b/c it was blocking some other things.

PR #352, opened 2020-04-04
Feature/module result cleanup

henry54809:


daq/runner.py
```diff
@@ -106,7 +110,8 @@ def get_run_info(self):
         return {
             'version': self._daq_version,
             'lsb': self._lsb_release,
-            'uname': self._sys_uname
+            'uname': self._sys_uname,
+            'daq_run_id': str(self._daq_run_id)
```
grafnu, line 114: I would do the str() on the initialization of the variable, rather than every place it's used.

PR #359, opened 2020-04-09
Refactoring host packet capture dumps.

grafnu:


daq/host.py
```diff
@@ -386,23 +386,28 @@ def _ping_test(self, src, dst, src_addr=None):
...
                                               intf_name=self._mirror_intf_name,
-                                              timeout=None, pcap_out=self._startup_file,
+                                              timeout=timeout, pcap_out=output_file,
                                               blocking=False)
-        self._tcp_monitor = helper
-        hangup = lambda: self._monitor_error(Exception('startup scan hangup'))
-        self.runner.monitor_stream('tcpdump', self._tcp_monitor.stream(),
-                                   self._tcp_monitor.next_line,
+        self._monitor_ref = helper
+        hangup = self._monitor_complete if timeout else (
```
henry54809, line 405: what if scan hangs up when there's a timeout? I think in your hangup function, if there's a timeout, you can check if timeout seconds has elapsed before calling monitor_complete.

Issue #313, opened 2020-03-20
Report formatting: increase space in the "notes" column in the report test result summary.


pisuke:

pisuke:
Addressed in #361.

PR #361, opened 2020-04-10
Increase notes column in report test

noursaidi:
#313 
Number of columns set to 1000 in to prevent pandoc assigning widths in the HTML precursor to the PDF reports thus allows the full width of the page to be used. If information to the left of the notes column (e.g. test names) increase, font sizes may additionally need to be reduced or other formatting change.

https://pandoc.org/MANUAL.html#tables

### DIFF
--- a/daq/report.py
+++ b/daq/report.py
@@ -176,7 +176,7 @@ class ReportGenerator:
         LOGGER.info('Generating HTML for writing pdf report...')
         pypandoc.convert_file(self.path, 'html',
                               outputfile=self._REPORT_TMP_HTML_PATH,
-                              extra_args=['-V', 'geometry:margin=1.5cm','--columns','1000'])
+                              extra_args=['-V', 'geometry:margin=1.5cm', '--columns', '1000'])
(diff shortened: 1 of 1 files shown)


pisuke:
@grafnu this is a quick one to fix the report table width issue with pandoc/weasyprint.
Tested, it passes on Travis and it works well functionally.

Not urgent, but we'll also address another warning that pops up from weasyprint in a different PR at some point:
```
WARNING:weasyprint:Expected a media type, got only/**/screen/**/and/**/(min-width: 480px)
WARNING:weasyprint:Invalid media type " only screen and (min-width: 480px) " the whole @media rule was ignored at 231:1.
WARNING:weasyprint:Expected a media type, got only/**/screen/**/and/**/(min-width: 768px)
WARNING:weasyprint:Invalid media type " only screen and (min-width: 768px) " the whole @media rule was ignored at 236:1.
```
but ignoring it for now as it doesn't prevent the PDF reports from being created in a readable format.

pisuke:
@grafnu just a quick reminder to merge this little bugfix for the report output.

PR #362, opened 2020-04-13
Update TLS methods and tests

jhughesbiot:
Resubmittal of work found in PR https://github.com/faucetsdn/daq/pull/317 which was closed due to too many merge conflicts.

Add TLS 1.0 support
Add RSA public key bit length validation
Add cipher key checks, warn only

Initial PR against grafnu for the TLS 1.3 additions: grafnu#269

Addresses Issues: #247 and #213


subset/security/Dockerfile.test_tls
```diff
@@ -1,13 +1,14 @@
-FROM daqf/aardvark:latest
+FROM daq/aardvark:latest
```
grafnu, line 1: should remain unchanged

subset/security/Dockerfile.test_tls
```diff
@@ -1,13 +1,14 @@
...
-RUN $AG update && $AG install openjdk-8-jdk git
+RUN $AG update && $AG install openjdk-11-jdk git
 
 COPY subset/switches/module_manifest.json .
 COPY subset/security/ .
 
 RUN cd tlstest && ./gradlew build
 
 CMD ["./test_tls"]
+
```
grafnu, line 14: extra line

testing/test_aux.out
```diff
@@ -11,12 +11,26 @@ RESULT info protocol.bacnet.version Protocol version: 1
...
+RESULT fail security.tls.v1_3.x509
+RESULT fail security.tls.v1_2
+RESULT fail security.tls.v1_2.x509
+RESULT fail security.tls.v1_3
+RESULT fail security.tls.v1_3.x509
+RESULT pass security.tls.v1
+RESULT pass security.tls.v1.x509
+RESULT pass security.tls.v1_2
+RESULT pass security.tls.v1_2.x509
+RESULT fail security.tls.v1_3
```
grafnu, line 28: there needs to be something other than just "fail" for the tests, otherwise there's no point in including them since they have no purpose (from a software perspective). How do you know these are working properly?

Shouldn't there also be a skip condition where there is to viable port to confirm TLS on?

jhughesbiot: The faux devices currently available do not have TLS 1.3 support.  I can't get passing results unless these devices are updated to support it.  All TLS version testing follow the exact same pattern but just defines the appropriate accepted TLS version for each test.

grafnu:
We need the faux device as a reference to know that it's testing the right
thing.

jhughesbiot:
I don't have any knowledge on the development of the faux devices and I haven't seen any documentation concerning that topic.  How do we go about adding TLS 1.3 support for them?

grafnu:
docker/Dockerfile.faux has all the dependencies (in docker container), and
at the end it runs misc/start_faux which actually executes the needful.
The best thing to do is find where the other tls things are managed and
iterate from there.  Searching the files for "tls" yields to be what I
believe are the relevant bits, so should provide a solid starting point.

jhughesbiot:
It appears to be contained as python code within the tls module here subset/security/tlsfaux/server.py.

There does not appear to be anything wrong in this code but the framework itself does not use a high enough version of pytho, openssl or something else I can't determine that would be required to utilize TLS 1.3.  Attempted to upgrade the faux dockerfile to java 11 where tls 1.3 is supported to see if that could resolve the issue but bacnet4j that also gets build in there which can't be built with this.  No idea what else would fail here but that's as far as it can bet. 

Supporting TLS 1.3 does not seem contained to this module and from my understanding of how this all works seems to exist in the overall framework which is currently not able to support it.  If i'm wrong, I need way more information on what a faux device is, how it works into the framework and how one develops it.

jhughesbiot:
I would really appreciate some feedback on getting the CI to pass as well as this has been an ongoing issue since the beginning and a constant struggle especially concerning the test_aux.

grafnu:
I assume you're looking at this test run output
<https://travis-ci.com/github/faucetsdn/daq/jobs/318986935>, or something
similar?

The only way to figure out what's going on is to look at them one by one
and see what the root cause is.  There's no real black magic, it's mostly
doing a diff against what is "expected", and it looks like a bunch of
things have changed, and that needs to be reflected appropriately.  The
expected output comes in two forms... one is the actual test log, and the
other is the (redacted) comparison against docs/device_report.md -- which
you have changed some of (but doesn't look like everything required).

E.g., looks like there's a bunch of differences like:

275c283
10158< security.tls.v1.2
10159---
10160> security.tls.v1_2

So, in one case it's using v1.2 and the other it's v1_2. First, I'd start
by cleaning up those things b/c if there's "real" issues underneath then we
need to find them.

If that's not what you're referring to, maybe can you be more specific?
Trying to tackle "all the things" doesn't work very well since it becomes a
fairly diffuse ask, in my experience.

As for the faux test -- looks like you're running into a dependency
quagmire, which isn't too surprising with trying to change something like
the Java version.  The way out of this, I think, is to create the ability
for different faux devices with different dependency sets... I'll put that
together and then we can sort out the v1.3 stuff.

The TLS stuff is entirely contained within the faux device and tls modules
-- there is nothing in the overall framework that knows anything about it.
If you have something specific to cite, please let me know and I can
follow-up. I'll explain the working model for that when I do the
alternate-faux (I'd like to write a proper document about it, rather than
just ad-hoc). But, again, let's focus on the CI test passing as-is before
addressing that bit.

Cheers,
  Trevor

jhughesbiot:
I would love to be more specific the problem is the travis log is I don't really know how you read any of that to determine exact failure points.  I updated the device_report to be consistent with the test_aux.out and still shows tons of differences.  Redacted is comparing to what?  What is actually wrong?

 It takes 30+ minutes for every change to be tested against aux and without knowing a true 1:1, this file needs to match this output and this output needs to match this file, this is becoming a very tedious task that's about 1000x longer than the actually development of the tests.

jhughesbiot:
I started looking into the test_aux.sh to see if this would help and found where the output files are being compared for the redacted files.  Manually looking I can see the differences and this seems to be an issue with the output file generation.  It is using a matching pattern somewhere when generating the results and adding wrong results to the report.

```
--------------------
security.tls.v1_3
--------------------
Verify the device supports TLS 1.3 (as a client)
--------------------
See log above
--------------------
RESULT skip security.tls.v1_3
RESULT fail security.tls.v1_3.x509
```

You can see that because the x509 has the same beginning, it's being added to the tls.v1_3 test section.  There's a lot of this going on so not sure where this problem needs to be corrected but it can't use a generic match and needs to be exact.

grafnu:
You are correct. I'm working onhttps://github.com/faucetsdn/daq/pull/367 to
fix this. The basics work (fix the problem you cite), just need to make
sure it doesn't break anything else.

grafnu:
Testing on Travis for development iteration is not an effective strategy.
It's slow, but that's how it works. You can run the tests locally, and then
each run is about 15 min, and you can also do the module testing in
isolation. It's better to build up from a solid foundation rather than
trying to rely on CI, which is intended to catch regressions and corner
cases.

THe bulk of the comparison is from the generated test output against
docs/device_report.md -- if you're running locally you can find the
generated output in out/redacted_file.md -- it is redacting things that
might change from run to run (dates, IP addresses, exact counts, etc...).
Usually though I just update docs/device_report.md directly, since it
should be clear what's expected to change.

jhughesbiot:
With the update now in master and now in a merge conflict, what is the preferred method to update a working branch to maintain my working code to get rid of this conflict and not get stuck in the same constant conflict state I was in before?

grafnu:
I don't understand how you're encountering merge conflicts... I don't know
exactly what you're doing and you seem to be encountering merge problems
that I do not (or maybe I just knew how to resolve them easily). Probably
need to be more explicit about exactly what you're doing and what you're
seeing, since the phrase "update now in master and now in a merge conflict"
doesn't detail enough about the situation for me to say anything
constructive. However, the intent with my PR was to avoid all of that and
get things back to a stable working state.

Did you try merging the PR I did against your branch?  From what I can
tell, it's a completely clean merge with no conflicts. The intent was you
could just merge that in and it should resolve any issues you've been
having.  That way, I did the merging on my end b/c it was easier for me
then to know exactly what was being done and what might need to be
resolved. When I look at the submitted PR
<https://github.com/jhughesbiot/daq/pull/5> it says that there are no
conflicts with the base branch.

jhughesbiot:
@grafnu 
Going to need some help on this CI issue.  No matter how many times i've tried, I can't get the aux test to pass and i'm constantly having to update to keep merge conflicts at bay.

grafnu:
It looks like my previous PR was never merged in?  That had cleaned up all
the CI tests before.  Anyway, I made a few tweaks and pushed the branch
again so they're again in PR #5 <https://github.com/jhughesbiot/daq/pull/5>.

grafnu:
Do you have a travis CI setup for testing these commits?  I was looking around for one and couldn't find it. I don't understand why you need to do these commits to fix aux because it should have been clear/green after the merge.  I'm wondering if you're environment isn't set up correctly which is why you're having so much trouble with this.  Essentially, I'm looking for a page like this, where your builds of various development branches happen. This build is from right after you merged my PR in, and it's showing passing... Were you able to see that somewhere?

jhughesbiot:
I have not setup my own travis CI to my fork and I have never had a successful build even with merging fixes.  I do not know what page you are referring where you saw successful build because it wasn't on this PR or branch.

There is no consistency between builds from my perspective.  My last build failed because it had additional messaging for the bacnet test 

RESULT skip protocol.bacnet.pic Bacnet device not found. **pics.csv not found in device type directory**

I removed the pics portion and now on this recent built it complains it's not there so...how do I ever hit this moving target?

There's also a ton of new differences now because the password test is failing again so more things no longer match.  This is nothing to do with my environment if the CI is so flaky it changes it's own results from build to build.

grafnu:
I see successful builds on my personal Travis:
https://travis-ci.org/github/grafnu/daq/builds/676470212

Flaxy tests do exist, and when I can find them I track them down.  The way
through that, however, is not to thrash about by tweaking the tests... but
rather to triangulate and/or reproduce by running the same build multiple
times. This is why I was asking if you could point me at a failed test run
that was exactly the same as what the PR should have yielded.  You should
be starting from the assertion that the PR I sent you (as shown in my PR)
is working (since that test run passed) -- and so if it doesn't work for
you then it's not something that should be tweaked in the output, but
rather something more fundamental going on that should be addressed.

This is why you really should set up your own, because it's a much more
controllable environment, and then you can simply point me at a build and
ask "why is this different?"... There should also be a build on the FaucetSDN
PR travis page <https://travis-ci.com/github/faucetsdn/daq/pull_requests>,
if you don't want to set up your own -- but you'd have to point me at which
one it is since it generates a new commit since it's testing the merge
rather than whatever is in your repo exactly.

jhughesbiot:
Let's say I had my environment setup with Travis on my own fork and this exact same PR builds perfectly.  I push it here and it fails (exactly what's going on right now) so I'm still stuck in a failed build that can't be accepted due to no fault of the PR.  The extra step of building elsewhere doesn't seem to solve anything.

Right now it's very clear my code is fine, the tests around it pass consistently but 2 other flaky modules are breaking this build.  This indicates a very clear failure of module isolation in the CI specifically around the aux test.  If there remains consistent flaky tests and modules, updating of anything not associated with them in anyway will continue to be blocked because of code that has nothing to do with those failures.  That is where this PR is and I do not know how to complete this PR with the system in this state.

grafnu:
It gives you something concrete you can point me at.  The problem with "it
doesn't work for me" is that I don't have enough information to actually do
anything with. If you can send me a link (like
https://travis-ci.org/github/grafnu/daq/builds/676470212) -- then I have
complete access to all the information necessary to try and figure out
what's going wrong. It's applying a rigorous methodology to getting to the
root problem here, rather than just thrashing around with tweaking things.
Also, it can help triangulate the problem -- e.g. given three different
builds (mine, yours, and faucetsdn) it will highlight the odd-man out which
gives much more information as to the root of the problem.  Right now I
only have one (mine) and everything else is somewhat abstract.

Alternatively, we know there is a problem somewhere.  And it's either with
your local code or setup, or with the CI code and setup.  The trick is to
partition it between your local stuff and the CI stuff -- otherwise it's
really difficult for me to help you. So, if you have the Travis CI setup
you can simply point me at the link and say "this doesn't work but it's the
exact same code as yours" -- which is isolating it as "not your problem."
Otherwise you might be trying to fix something that you can't reasonably
address.

jhughesbiot:
I don't understand the issue you have to determine the failure point.  All code changes are in this PR, they can be run locally for validation.  They are run here against this project where you can see all the code that was and wasn't changed for the CI to build.  There is no changes to anything in the CI except for the output files to be checked against.  If two builds from this CI with the exact same code have different results, and that's what the issue is here, this makes no difference if I ran it on another CI with the exact same code.  

This PR has now gotten to the point where I have to help diagnose other flaky tests to get it accepted that have nothing to do with my code changes.  It is clear from the build failures the bacnet pics and the password tests are not consistent but are causing build failures.  This has nothing to do with the security module but this PR is now being put on the hook for failures of other modules?  There's plenty of other PR's I can see with build failures being accepted probably overlooking the same bad module results so what needs to happen to get past this.

I have no reason to believe anything will come of creating another Travis CI build and installation when I already know where the failure lies and it's not in the code from this PR.

jhughesbiot:
@grafnu 
How can I trigger a new build of the same code to get over this test flakiness issue until I get a valid build?

grafnu:
I don't understand your question -- do you mean you want to rebuild your
docker images?

jhughesbiot:
No I mean for this PR, is there a way to trigger a new build to try without making any changes to the code to trigger it.  The current failed build is due to the password test which fails sometimes and others not.  The previous build failed because the bacnet pics messaging was different.  I updated it to what it's supposed to be but now the password test fails again.  There's nothing wrong with the code, just that the flaky tests are causing the failure.  If I can trigger new builds until I get lucky, then this PR should be able to be merged and closed.

grafnu:
oh, you mean the Travis run?  Yes... I just restarted it.  For your own
Travis setup you can do it as well... but for a PR run it requires special
permissions.  I'll try to remember to take a look to see if it happens
again... but if it's truely flaky then we don't *have* to have all the
tests pass... that's just for sanity sake.

testing/test_aux.out
grafnu: Sill missing the pass variants for v1_3... the java-11 version of the faux device is available so should be suitable for adding in 1.3

testing/test_modules.out
```diff
@@ -5,14 +5,47 @@ RESULT skip base.startup.dhcp Using static ip 10.20.0.5
...
-RESULT skip security.tls.v3
-RESULT skip security.x509
+RESULT skip security.tls.v1
+RESULT fail security.tls.v1.x509
+RESULT skip security.tls.v1_2
+RESULT fail security.tls.v1_2.x509
+RESULT skip security.tls.v1_3
+RESULT fail security.tls.v1_3.x509
+RESULT fail security.tls.v1.x509
+RESULT skip security.tls.v1_2
```
grafnu, line 15: results are being duplicated still?  This is weird because when I run test_modules.sh locally I don't see the duplication.  Anyway... I'll investigate further but I wouldn't consider it a blocker for your PR since I don't think it's the tls-test code!

grafnu:
Ok, after a few tries everything passed.  What's weird is Haoli had a PR in
that ran successfully for 3x times in a row. We're working on resolving the
underlying cause of the flakiness, so hopefully that'll be fixed soon.  I'm
doing another review of the PR now.

testing/test_modules.out
jhughesbiot: Not sure what duplication you're seeing, this looks just like the 3 skip/pass/fail scenarios. I still don't really know how the faux devices are built.  There probably needs to be some verification that TLS 1.3 is actually available in them separate to the test to rule out which one is the issue.

grafnu:
The result

RESULT skip security.tls.v1_2

is in there twice.

jhughesbiot:
I was looking at aux so wrong file.  I updated the tests to handle skip properly for x509 tests and ran it locally and don't see the duplication.  Going to update that file with the tests without the duplication and see what the results look like.

jhughesbiot:
So it appears the duplicates only show up when running in travis but not running the module_test locally.

grafnu:
I have discovered this as well... I'm looking into that, because for sure
they should be consistent!

jhughesbiot:
Well at least for now the module test is fixed but aux is still having that BACnet inconsistency error.  Still not sure how to handle the TLS 1.3.  The test does check it the same way it does for 1.1. and 1.2 but the faux devices don't support it so they will never have a pass result until that happens.

grafnu:
From a software perspective (my role) we need a reference design for the
test to operate against, otherwise it's not complete. So, that gives us
three options:

   1. We take out the 1.3 test for now.
   2. Include the necessary capability to get it done.
   3. Up-level to PM to address the requirement.

grafnu:
Found the problem with the results repeating -- the docker system is pulling down an older docker image from DockerHub for something, so it's relying on some old code in some circumstances.  I'm working on a fix now (after I figure out what it should be!).  So, your local setup is likely behaving correctly... it's Travis that's misbahving.

grafnu:
Ok, this has been fixed in master now... as per
https://travis-ci.org/github/grafnu/daq/builds/679265823 I did a test merge
and it's working in that there's no duplicates. It'll just require a minor
tweak to the .out file to remove them there for a match.

jhughesbiot:
I think I found the issue, TLS 1.3 support was introduced in python 3.7 and the current faux containers are at 2.7.17.  Not sure if updating the core aardvark image would be the better way or if this would cause issues to others.  Either way, faux2 will need to get updated to 3.7.

I think maybe remove 1.3 from this PR and create a separate 1.3 support add might be the way to go if changing python version seems like it might be a can of worms.  Thoughts?

grafnu:
Ok -- I updated faux2 to be python3.7 -- literally a three character
change!  https://github.com/faucetsdn/daq/pull/410/files  -- if you just
want to edit your file directly.  I think the TLS is the only stuff that
uses it, so nothing else would break.  You can install 3.7 and 2.x
side-by-side, but didn't bother with that.

jhughesbiot:
If I check the version with python --version it doesn't resolve python after this change.  If I use python3.7 --version, it does resolve.  But since nothing outside of the container would tell other commands where to resolve python, and I haven't been able to follow the logic that actually starts the tls faux device, i'm not even sure how it still works.  The test_module code does still work but no change to the output so i'm not sure really where the python version comes into play on the tls testing.

I assumed something was calling this file https://github.com/faucetsdn/daq/blob/master/subset/security/tlsfaux/server.py that was responsible but I can't seem to find any reference to anything invoking it so I really don't know what these faux devices are doing to create valid TLS versions.

jhughesbiot:
On further review with this change, it doesn't appear that the python version within the faux container was enough.  While trying to track down how this test was actually working, I found that the nginx file used in the tls passing test didn't define tls 1.3.  Once I updated that file to include 1.3, the tests started passing.  I reverted the python version back just to confirm the python upgrade wasn't still necessary and it passes at python 2 so even though it's not required, can't hurt to keep it updated.

grafnu:
OK, well I at least fixed the python problem. In the faux2 container now it
maps 'python' to 'python3.7' -- which wasn't there before. Also adds some
diagnostics to the initial output.

jhughesbiot:
Well it looks like we have the 3 results now for all the tls versions but the aux test is failing.  I'm looking at the out compared to what it's generated and the report it's coming up with makes no sense.  It's like a spacing issue that doesn't exist in anyway that I can find.

grafnu:
Link please.

jhughesbiot:
https://travis-ci.com/github/faucetsdn/daq/jobs/330733626

grafnu:
So the space changes are something that are in a change in the docs/device_report.md file -- I'm assuming not intentional on your part, but there nonetheless. There's a change in trailing white-space on the first one, and on the second the spaces were replaced with tabs. Two options -- either manually edit the file to remove those specific changes, or we figure out how they got there in the first place and fix the root cause.

How, exactly, did you generate that file?  It seems to be a redacted version of it, which is in itself fine... but how did you generate it? I'm trying to understand where the unintended changes came from because if it was a straight file-copy they wouldn't be there. Some command/step/process in the flow must have been trying to change things.

jhughesbiot:
At some point copied the existing docs in the build and modified them directly to the test changes because I couldn't figure out what was different and the issues still remain so how they got different makes no sense to me.

I just copied the device_report again and then added all my changes again and that seems to have fixed the tab/spaces issue with just the flaky bacnet test showing up on the last build.

grafnu:
k -- if it's just the flaky bacnet test then we're likely good to go on
that front -- that doesn't need to be resolved.

jhughesbiot:
Sounds good.  I think everything else has been resolved as far as feedback on this PR.

PR #363, opened 2020-04-14
UDMI Gateway Schema Update

grafnu:


jcbrough:
I'm thinking of this and trying to keep the amount of setup on the cloud side to a minimum. The IoT-Core registry and gateway have to be created manually as the immutable data that they contain cannot be populated by the gcloud sdk (i.e. the gateway config data that contains the list od devices that the gateway is acting for).

It is then possible for the gateway to create the IoT-Core devices, but the immutable data in the device config blocks cannot be created by the gcloud SDK so my proposal is that this data is just in the metadata block and the gateway publishes this to the device 'state' topic after creation

grafnu:
John, we're unlikely to want the gateway to be able to create/configure
devices directly, that's not its function and would require granting
excessive permissions to an "untrusted" node in the system.  Rather, I
think what we should do is update the registrar tool to properly configure
and setup the components.  So, the site config would then have an
appropriate model, and the tool would configure things automatically.
Maybe functionally equivalent to what you're describing, but rather than
the gateway doing anything, it's done from the registrar tool.

However, all of that is just automation of the underlying process -- but
everything *should *be able to be done manually, even if it's tedious.
 Are you suggesting/highlighting a change that needs to be made in the
specification (the JSON files and schemas), or are you talking about the
*process* around how everything is setup and configured?

jcbrough:
That's fine Trevor, I was just trying to minimize the amount of re-entering of the same data that would otherwise be required. From the device side all of the mapping of data from the proxied devices to the gateway has to happen anyway and this is isolated from the cloud functions. From what I have built so far from the example at https://cloud.google.com/iot/docs/how-tos/gateways/mqtt-demo it looked like it was p;ossible to then get the gateway to auto-populate a lot of this data.

But I take your point - just because we do it, it doesn't mean that we want to. If only from a security protection perspective.

Maybe we can utilize a common schema though so that only needs defining once.

grafnu:
"Maybe we can utilize a common schema though so that only needs defining
once."  ==> Yes, 100% this.  That's exactly what UDMI is too, a set of data
sets and schemas, that is mostly separated out form mechanism (how the
data gets into various places).

If there's some specific piece of information you're looking to represent,
please let me know and we can sort out where it should go.

jcbrough:
Further to some code development to breadboard this out I have come up with an overview as to how we might provision these gateways:

- We would need a JSON block to tell the gateway what cloud parameters to use (hostname/tcp port/cloud region/gcloud project/gcp registry/gcp gateway). This would be defined by the MSI and would be stored in a suitable BAS object (e.g. CSV object in BACnet), that is loaded into the controller during commissioning on premises by the SI

- Once it knows the cloud parameters the gateway device can auto-create the gateway in the gcp project, the public/private key pair and then write the public key into the gateway device in the gcp.

- The MSI would then populate the config block for the gateway device, which contains the list of proxy devices, create the proxy devices themselves and their config blocks - which contain the immutable data for the individual devices and also the details of the point mapping. Then 'bind' these proxy devices to the gateway. This would be accomplished by the MSI using the 'registrar tool'

- Once everything is defined the phyical gateway on premises will receive the config state block from the gateway, which tells it the devices it has to proxy for and for each proxy device it will receive a device config block, which gives it the mapping data that it needs to know in order to publish the data

Comments please...

grafnu:
Architecturally, there's three things here that should provide the
framework for what you're talking about:

   - *Model* -- the "json description" of how everything should work. This
   is, ideally, something that is provided a-priori b/c it's all known up
   front (no surprises), but there might be a *separate* process of
   *discovery* that fills this in from discover information -- but this is
   conceptually different than a gateway.
   - *Registrar* -- a tool that does any of the privileged work in the
   cloud, e.g. adding/removing/updating devices. This is something that is
   either run by a person (the MSI), or from a cloud based service... but
   *not* the gateway, which is Gateway -- the thing on-prem is generally
   considered only to be "semi-trusted"
   - *Gateway* -- the thing on-prem that proxies data for other devices. It
   may also send discovery information like "I saw this device, but I didn't
   know what it was" -- and that information could be consumed by the
   cloud-side to update the registry/config/model as appropriate.

The key thing here is recognizing that the gateway will *not* have the
authority to manipulate the registry, since that would be a big security
problem, and is not required for normal on-going operation.

Some of the stuff you site (GCP project, registry, IP address, etc...) --
should be in there already somewhere, since all of that is required for any
system communicating directly with the cloud, independent of the gateway
concept.

grafnu:
OK, this has been updated with the Registrar tool that can now read in the model and configure Cloud IoT appropriately. Also the Pubber client has been updated to use the configured gateway. This also takes care of the initial "config" block, so that part should have an out-of-the-box working example as well.

PR #366, opened 2020-04-14
Create alternate faux container image

grafnu:


bin/docker_build
```diff
@@ -41,7 +41,8 @@ for file in $docker_files; do
     tag=daqf/$module
     found=$(docker images $tag 2>/dev/null | grep $module) || true
     if [ -n "$DAQ_TARGETS" -a -z "${test_set[$module]}" ]; then
-        echo Skipping non-enabled $tag
+        #echo Skipping non-enabled $tag
```
henry54809, line 44: Is it intended to be commented out?

grafnu: Yes -- it was very spammy in the output. I was leaving it in as a comment with a pseudo output (obvious in case somebody wanted to enable/debug)

PR #367, opened 2020-04-14
Fix result extraction for reports

grafnu:


jhughesbiot:
The update looks ok but it appears the aux test is broken on the build but not related to this PR.  Probably should get that addressed so it can be fully validated before merging.

grafnu:
Ok, should all be fixed now... the tests are passing.

PR #369, opened 2020-04-16
Feature/proper ip test

henry54809:


daq/host.py
```diff
@@ -47,16 +48,31 @@ class MODE:
...
 
+def _ip_change_verify(host):
+    if len(host.all_ips) <= 1:
+        return None
+    return len(set(map(lambda ip: ip["ip"], host.all_ips))) > 1
+
+_DEFERRABLE_CHECKS = [
+    {
+        "name": "ipchange",
+        "action": lambda host: host.gateway.execute_script('new_ip', host.target_mac),
```
grafnu, line 59: can you create a helper function for this so it's like host.gateway.set_new_ip(host.target_mac) or similar?  I'd rather not expose the mechanism like this.

daq/host.py
```diff
@@ -47,16 +48,31 @@ class MODE:
...
+        "verify": _ip_change_verify
+    }
+]
 
 def pre_states():
     """Return pre-test states for basic operation"""
     return ['startup', 'sanity', 'ipaddr', 'base', 'monitor']
 
+def deferrable_check_states():
```
grafnu, line 68: how is deferrable different than post?

Also, I think ipchange would need to go near the front of the sequence, since part of what we need to check is that everything else still works after the IP address has changed.  E.g., it's possible (and I've seen it), that the IP stack of the device works, so it responds ot ping -- but an application stack like bacnet will fail (since it cached the original IP address).

daq/host.py
```diff
@@ -116,6 +133,10 @@ def __init__(self, runner, gateway, target, config):
         self._trigger_path = None
         self._startup_file = None
         self.timeout_handler = self._aux_module_timeout_handler
+        self._deferrable_checks = list(filter(lambda check: self._test_enabled(check["name"]),\
```
grafnu, line 136: avoid line continuation

daq/host.py
```diff
@@ -300,20 +321,49 @@ def _main_module_timeout_handler(self):
...
+                self._check_module_timeout(self.test_name, check["test_start"])
+                try:
+                    result = handler(self)
+                    if result is not None:
+                        self.record_result(name, state=(MODE.DONE if result else MODE.MERR))
+                        check["done"] = True
+                except Exception as e:
+                    self.record_result(name, exception=e)
+                    self._report.accumulate(self.test_name, {ResultType.EXCEPTION: str(e)})
+            self._active_deferrable_checks = list(filter(lambda check: "done" not in check, \
```
grafnu, line 358: avoid continution

grafnu, line 358: what is this line doing?  The problem with the list(fliter( stuff, although more concise, is it's often very obtuse...

daq/host.py
```diff
@@ -364,13 +414,15 @@ def trigger(self, state=MODE.DONE, target_ip=None, exception=None, delta_sec=-1)
         self._trigger_path = os.path.join(self.scan_base, 'ip_triggers.txt')
         with open(self._trigger_path, 'a') as output_stream:
             output_stream.write('%s %s %d\n' % (target_ip, state, delta_sec))
+        self.all_ips.append({"ip": target_ip, "timestamp": gcp.get_timestamp()})
         if self.target_ip:
-            LOGGER.debug('Target port %d already triggered', self.target_port)
-            assert self.target_ip == target_ip, "target_ip mismatch"
+            LOGGER.debug('Target port %d already triggered' % self.target_port)
```
grafnu, line 419: should be , not % -- you should be getting a lint warning for this, no?

daq/host.py
```diff
@@ -470,8 +522,12 @@ def _monitor_complete(self):
 
     def _monitor_continue(self):
         self._state_transition(_STATE.NEXT, _STATE.MONITOR)
+        self._execute_deferrable_check_actions()
```
grafnu, line 525: i'm still horribly confused by the use of the word "deferrable" here... I'm not sure that's the right word. If you could describe what it's doing in a sentence maybe I could suggest an alternative.

daq/host.py
```diff
@@ -570,17 +634,31 @@ def _host_dir_path(self):
...
         return os.path.join(self._host_dir_path(), 'tmp')
 
+    def _exists_new_ip_assignment(self, timestamp):
+        test_start = gcp.parse_timestamp(timestamp)
+        b_fn = lambda ip: gcp.parse_timestamp(ip["timestamp"]) < test_start
+        a_fn = lambda ip: gcp.parse_timestamp(ip["timestamp"]) >= test_start
+        before, after = list(filter(b_fn, self.all_ips)), list(filter(a_fn, self.all_ips))
+        return len(after) > 0 and after[-1]["ip"] != before[-1]["ip"]
+
+
```
grafnu, line 644: extra blank line

daq/host.py
```diff
@@ -570,17 +634,31 @@ def _host_dir_path(self):
...
-        if failed and self._fail_hook:
-            fail_file = self._FAIL_BASE_FORMAT % host_name
-            LOGGER.warning('Executing fail_hook: %s %s', self._fail_hook, fail_file)
-            os.system('%s %s 2>&1 > %s.out' % (self._fail_hook, fail_file, fail_file))
+        if failed:
+            if self._fail_hook:
+                fail_file = self._FAIL_BASE_FORMAT % host_name
+                LOGGER.warning('Executing fail_hook: %s %s', self._fail_hook, fail_file)
+                os.system('%s %s 2>&1 > %s.out' % (self._fail_hook, fail_file, fail_file))
+            # Check if new ip was assigned during test run and retry
```
grafnu, line 657: this seems non-deterministic... how does this work, then? It's just letting the change happen during some arbitrary test, and then restarting?  I'm not sold on that's the right thing.... seems like we might want it to be a more explicit phase...

misc/module_config.json
```diff
@@ -3,6 +3,9 @@
     "ipaddr": {
       "timeout_sec": 300
     },
+    "ipchange": {
```
grafnu, line 6: I don't know how to comment on it -- but there's an extra flie "Blank Diagram.png"?

misc/networking_scripts/new_ip
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash -e
+#
+# Forces a new ip on a DHCP device
+LEASES_FILE=/var/lib/misc/dnsmasq.leases
+mac_addr=$1
+if [ -z "$mac_addr" ]; then
+    echo "Usage: ./new_ip mac_addr"
```
grafnu, line 7: no need for ./ in usage statement

subset/connection/test_macoui
```diff
@@ -17,7 +17,7 @@ java -jar mac_oui/build/libs/mac_oui-all.jar $TARGET_MAC
 
 RESULT_AND_SUMMARY="$(grep "RESULT" $LOCAL_REPORT)"
 grep -v "RESULT" $LOCAL_REPORT >> $REDACTED_LOG
-sleep 5
+sleep 10
```
grafnu, line 20: did this do anything?  I"ve been sorting through why that test is flaky -- still don't know yet... but I wouldn't have guessed this would make any difference.

subset/pentests/test_nmap
```diff
@@ -103,3 +103,4 @@ write_out_result $REPORT \
                  "$TEST_DESCRIPTION" \
                  "$(cat $REDACTED_LOG)" \
                  "$RESULT_AND_SUMMARY"
+
```
grafnu, line 106: extra line

testing/test_dhcp.out
```diff
@@ -5,5 +5,5 @@ DHCP Tests
 02: ['02:ipaddr:TimeoutError']
 Device 1 ip triggers: 1 0
 Device 2 ip triggers: 0 0
-Device 3 ip triggers: 0 1
+Device 3 ip triggers: 1 1
```
grafnu, line 8: I feel like we should have a more explicit test for this... this seems very indirect/hidden.  The ip_triggers file has a list of the ip addresses used, so seems like we should be able to awk|uniq that file to make sure there's ==2 different addresses?

daq/host.py
henry54809: if the intention is to check whether a new ip has an effect on the device, shouldn't we run it once with one ip, wait for a new ip, then run the test again? I think we can define a set of things a device can ask for when running a test and that can include a new ip, multi dhcp etc through some api on the gateway. When the test is over, gateway can return to default configurations. It'll be easier to do that than making the existing DAQ workflow more complicated.

henry54809:
@grafnu PTAL . Travis pending.

PR #371, opened 2020-04-17
Docs update

henry54809:


docs/site_path.md
```diff
@@ -25,6 +25,8 @@ The merged config file is written to `{site_path}/mac_addrs/{mac_addr}/aux/modul
 and is available to a test container at runtime (along with all the other files in the `aux`
 directory).
 
+![Diagram](module_config.png)
```
grafnu, line 28: What's this? I don't see the file (missing commit?)... also hesitant to have images in there if it's text of some kind... ?

henry54809: the picture is there under docs in this commit. 2 files above this one. github pulls in the image when you use that syntax. I've previewed the markdown files to make sure it works.

grafnu: oh right. was just somewhat hidden... found it.

PR #373, opened 2020-04-20
Base configurator unit test

anurag6:
The test creates a temp config file from an object and reads it using the configurator module. It tests whether the read object is same as the original object.


testing/run_unit_tests.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
```
grafnu, line 1: This should be called from one of the CI test scripts (travis) -- I'd recommend testing/test_modules.sh -- even though it's not modules (will rename later).

testing/run_unit_tests.sh
```diff
@@ -0,0 +1,12 @@
...
+
+MINCOVERAGE=92
+SCRIPTPATH=$(readlink -f "$0")
+TESTDIR=`dirname $SCRIPTPATH`
+BASEDIR=`readlink -f $TESTDIR/..`
+cd $BASEDIR || exit 1
```
grafnu, line 7: all the || exit 1 start to get ugly... I'd just add -e to the top and then explicitly allow whatever commands are allowed to fail.

testing/run_unit_tests.sh
```diff
@@ -0,0 +1,12 @@
...
+
+$BASEDIR/venv/bin/coverage erase || exit 1
+PYTHONPATH=$BASEDIR/daq $BASEDIR/venv/bin/coverage run --source $BASEDIR/daq -m unittest discover -s $TESTDIR/unit/ -p "test_*.py" || exit 1
```
grafnu, line 10: add a line-break in here somewhere for readability

testing/unit/test_configurator.py
```diff
@@ -0,0 +1,47 @@
+"""Unit tests for configurator"""
+
+import unittest
+import os
+
+from daq.configurator import Configurator, print_config
+
+def dict_to_str(obj):
```
grafnu, line 8: dict_to_conf_str -- it's not just a "str" which could be anything... it's more specific than that!

testing/unit/test_configurator.py
```diff
@@ -0,0 +1,47 @@
...
+        'default_timeout_sec' : '350',
+        'base_conf' : 'misc/module_config.json',
+        'site_path' : 'local/site/',
+        'initial_dhcp_lease_time' : '120s',
+        'dhcp_lease_time' : '500s',
+        'long_dhcp_response_sec' : '105'
+    }
+
+    def setUp(self):
+        tmpfile = open('temp.conf', 'w+')
```
grafnu, line 32: extract "temp.conf" to a constant variable since it's used in 3x places

testing/unit/test_configurator.py
```diff
@@ -0,0 +1,47 @@
...
+        tmpfile.close()
+
+    def tearDown(self):
+        os.remove('temp.conf')
+
+    def test_config_load(self):
+        """Test config is loaded properly"""
+        configurator = Configurator()
+        args = ['test', 'temp.conf']
+        read_config = configurator.parse_args(args)
```
grafnu, line 43: I hate English: red == read != read

testing/run_unit_tests.sh
anurag6: There already exists a test_modules.sh which is called. test_units?

testing/unit/test_configurator.py
anurag6: I am not sure I understand, Could you elaborate.

testing/run_unit_tests.sh
anurag6: Addressed.

testing/run_unit_tests.sh
anurag6: Addressed.

testing/unit/test_configurator.py
anurag6: Addressed.

testing/unit/test_configurator.py
anurag6: Addressed.

testing/unit/test_configurator.py
grafnu: "Read" past tense is different than "read" present tense.  I was going to make a comment about how you should use "read" instead of "read"... and then I hate english.

testing/run_unit_tests.sh
grafnu: it should go in test_modules... there's a high overhead for a new top-level test, so shouldn't create new ones unless absolutely necessary.

testing/run_unit_tests.sh
grafnu: Except for the "exit 1"?

PR #383, opened 2020-04-23
Feature/ui improvements

henry54809:
1. Adding reports and module configs for all non docker modules and upload them to gcp
2. Adding links for all uploaded files to dashboard
3. Better usability for run time module config modification.

### DIFF
--- a/daq/daq.py
+++ b/daq/daq.py
@@ -17,7 +17,7 @@ LOGGER = logger.get_logger('daq')
 ALT_LOG = logger.get_logger('mininet')
 
 _PID_FILE = 'inst/daq.pid'
-
+_LOG_FORMAT = "%(levelname)-8s%(module)-15s%(asctime)-25s%(message)s"
 
 class DAQ:
     """Wrapper class for configuration management"""
@@ -33,7 +33,7 @@ class DAQ:
         daq_env = config.get('daq_loglevel', log_def)
         level = minilog.LEVELS.get(daq_env, minilog.LEVELS['info'])
 
-        logger.set_config(level=level)
+        logger.set_config(level=level, format=_LOG_FORMAT)
 
         # For some reason this is necessary for travis.ci
         ROOT_LOG.setLevel(level)


daq/host.py
```diff
@@ -325,12 +324,22 @@ def terminate(self, reason, trigger=True):
...
         self._state_transition(_STATE.TERM)
-        self.record_result(self.test_name, state=MODE.TERM)
         self._monitor_cleanup()
         self.runner.network.delete_mirror_interface(self.target_port)
+        self._report.finalize()
+        json_path = self._report.path + ".json"
+        with open(json_path, 'w') as json_file:
+            json.dump(self._report.get_all_results(), json_file)
+        remote_paths = {}
+        upload_path = self._get_unique_upload_path(self._report.path)
```
grafnu, line 334: I would create a helper function here... I'm seeing 3x uses of the exact same code pattern

daq/host.py
```diff
@@ -595,28 +594,27 @@ def _docker_callback(self, return_code=None, exception=None):
...
         activation_log_path = os.path.join(self._host_dir_path(), 'activate.log')
         module_config_path = os.path.join(self._host_tmp_path(), self._MODULE_CONFIG)
+        remote_paths = {}
         for result_type, path in ((ResultType.REPORT_PATH, report_path),
                                   (ResultType.ACTIVATION_LOG_PATH, activation_log_path),
                                   (ResultType.MODULE_CONFIG_PATH, module_config_path)):
             if os.path.isfile(path):
                 self._report.accumulate(self.test_name, {result_type: path})
-                self._gcp.upload_file(path, self._get_unique_upload_path(path))
+                remote_path = self._gcp.upload_file(path, self._get_unique_upload_path(path))
```
grafnu, line 606: 4x uses?

daq/host.py
```diff
@@ -595,28 +594,27 @@ def _docker_callback(self, return_code=None, exception=None):
...
+                remote_paths[result_type.value] = remote_path
+        self.record_result(self.test_name, state=state, code=return_code, exception=exception, \
```
grafnu, line 608: no line continuation necessary

daq/host.py
```diff
@@ -658,7 +661,7 @@ def clear_port(gcp_instance, port):
         }
         gcp_instance.publish_message('daq_runner', 'test_result', result)
 
-    def _record_result(self, name, run_info=True, current=None, **kwargs):
+    def _gcp_record_result(self, name, run_info=True, current=None, **kwargs):
```
grafnu, line 664: why is this now specific to gcp?  I think it's bleeding the abstraction... this would be the inflection point for any non-gcp recording as well...

PR #395, opened 2020-05-01
Feature/ui improvements

henry54809:
1. Adding Filters: runid, port, device id, origin, site and test date.
2. Fixing data loss and data inconsistencies on cloud functions.


daq/host.py
```diff
@@ -647,18 +647,6 @@ def record_result(self, name, **kwargs):
                 self._report.accumulate(name, {ResultType.RETURN_CODE: kwargs["code"]})
             self._report.accumulate(name, {ResultType.MODULE_CONFIG: self._loaded_config})
 
-    @staticmethod
```
grafnu, line 650: where is this functionality going?

firebase/functions/index.js
```diff
@@ -4,158 +4,132 @@
...
-  });
-  rundoc.delete().catch((error) => {
-    console.error('Error deleting', port, runid, error);
-  });
-}
-
 exports.daq_firestore = functions.pubsub.topic('daq_runner').onPublish((event) => {
   const message = event.json;
   const origin = event.attributes.origin;
-  const site_name = event.attributes.site_name || 'unknown';
```
grafnu, line 34: this refactoring (changing the names of things) needs to be done independently of any functionality change -- otherwise it's near impossible to do an effective code review.

firebase/functions/index.js
```diff
@@ -4,158 +4,132 @@
...
-const {PubSub} = require(`@google-cloud/pubsub`);
+const { PubSub } = require(`@google-cloud/pubsub`);
 const pubsub = new PubSub();
 
-const EXPIRY_MS = 1000 * 60 * 60 * 24;
-
 admin.initializeApp(functions.config().firebase);
 const db = admin.firestore();
 
-function deleteRun(port, port_doc, runid) {
```
grafnu, line 15: where is this functionality going?

henry54809: It's not needed so I deleted it. It's the same with clearport function.

grafnu: Both the deleterun and clearport existed for a reason, so not sure if the functionality has been moved elsewhere, or has been obsoleted in some way?

henry54809: Yes clearport result is now obsolete with my changes and there's no need to delete any previous run data with the way the data is queried/ structured.

grafnu:
Another solution I had to the "should stuff be deleted" conundrum is to
make it configurable... Then we can deal with the implications of it later
more easily.

henry54809:
PTAL

firebase/public/main.js
```diff
@@ -387,6 +389,31 @@ function displayVersions(originId) {
     document.querySelector('#lsb-version').innerHTML = message.lsb;
     document.querySelector('#sys-version').innerHTML = message.uname;
     document.querySelector('#daq-versions').classList.add('valid');
+    if (message.timestamp && message.timestamp < heartbeatTimestamp) {
```
grafnu, line 392: this code doesn't seem related to the encapsulating function (displayVersions)... I'd recommend extracting a bunch of this to a separate function, and then maybe adjusting the names to match what their role is?

henry54809:
reorganized function. PATL

henry54809:
PTAL.

daq/runner.py
```diff
@@ -108,12 +109,17 @@ def _send_heartbeat(self):
...
     def get_run_info(self):
         """Return basic run info dict"""
-        return {
+        info = {
             'version': self._daq_version,
             'lsb': self._lsb_release,
             'uname': self._sys_uname,
             'daq_run_id': str(self._daq_run_id)
         }
+        data_retention_days = float(self.config.get('run_data_retention_days'))
```
grafnu, line 118: I think the default needs to be set here to catch the cases of a misconfigured system still expiring data. For now, at least, I want to keep that as the default behavior and relying on the default system.conf file is very error prone.

firebase/functions/index.js
```diff
@@ -55,12 +121,20 @@ function handleTestResult(origin, siteName, message) {
...
+  const updates = [
     originDoc.set({ 'updated': timestamp }),
     siteDoc.set({ 'updated': timestamp }),
     portDoc.set({ 'updated': timestamp }),
     deviceDoc.set({ 'updated': timestamp })
-  ]).then(() => {
+  ];
+  if (message.config && message.config.run_info
+    && message.config.run_info.expiration) {
+    const expirationDoc = originDoc.collection('expiration').doc(message.runid);
```
grafnu, line 132: I think the DB schema has gotten more complicated with some of these changes, and it's getting to the point where I think we should document it since there's getting to be more and more dependencies between various parts of the setup.  I don't inherently understand why this expiration time can't be a property of the data itself -- I'm sure there's a good reason for it, but something we'll have to make more explicit, I think.

daq/runner.py
```diff
@@ -115,9 +115,9 @@ def get_run_info(self):
             'uname': self._sys_uname,
             'daq_run_id': str(self._daq_run_id)
         }
-        data_retention_days = float(self.config.get('run_data_retention_days'))
+        data_retention_days = self.config.get('run_data_retention_days', None)
```
grafnu, line 118: There's no actual default here -- should be some retention value.  Behavior should be initially similar to what it was before.

PR #399, opened 2020-05-04
add checks for interfaces to switch being defined and up

pbatta:


bin/physical_sec
```diff
@@ -23,12 +23,22 @@ if [ -z "$ext_ctrl" ]; then
     false
 fi
 
+if ! ip -br link show dev "$ext_ctrl" | fgrep "UP" &> /dev/null; then
+    echo Error checking ext_ctrl interface [$ext_ctrl] status is UP, for physical_sec
```
grafnu, line 27: nit: The error is not in the checking -- the check results in the error. I'd recommend something more direct:

Error: ext_ctrl interface [$ext_ctrl] is not IP

(and below)

pbatta: addressed. thanks.

PR #405, opened 2020-05-06
add script to collect diagnostic info by user before they reach out f…

pbatta:
…or technical support

### DIFF
--- a/bin/techsupport
+++ b/bin/techsupport
@@ -1,18 +1,35 @@
 #!/bin/bash -e
 
-TMPDIR=./techsupport
+ROOT=$(realpath $(dirname $0)/..)
+cd $ROOT
+
+TMPDIR=/tmp/techsupport
 mkdir -p $TMPDIR
 
 echo Collecting system information
-sudo ovs-vsctl show > $TMPDIR/ovs_vsctl_show.txt
-lsb_release -a > $TMPDIR/lsb_release.txt 2>/dev/null
-ip route > $TMPDIR/ip_route.txt
-ip link show > $TMPDIR/ip_link_show.txt
-arp -n > $TMPDIR/arp.txt
+function save_to_file {
+  cmd=$1
+  outfile=${cmd// /_} # cleaner form of filename
+  outfile=${outfile//\-/}
+  $1 > $TMPDIR/$outfile.txt 2>/dev/null
+}
+save_to_file "sudo ovs-vsctl show"
+save_to_file "lsb_release -a"
+save_to_file "ip route"
+save_to_file "ip link show"
+save_to_file "arp -n"
+save_to_file "git describe --dirty"
+save_to_file "git status"
+save_to_file "git remote -v"
+save_to_file "git log -n5"
 
 if [ -d inst ]; then
-    echo Collecting run logs and config from 'inst' and 'local'
+    echo Collecting runtime info from 'inst'
     cp -r inst $TMPDIR
+fi
+
+if [ -d local ]; then
+    echo Collecting config from 'local'
     cp -r local $TMPDIR
 fi
 
@@ -23,13 +40,21 @@ fi
 # If an external switch was set up, try pinging it and saving that capture
 if [ -n "$ext_ctrl" ] && [ -n "$ext_addr" ]; then
     echo Saving packet capture of ping to external switch
-    sudo timeout 3s tcpdump -i $ext_ctrl -w $TMPDIR/ping_switch.pcap &>/dev/null &
-    sleep 1 # give a second for tcpdump to settle
-    sudo arp -d $ext_addr
-    ping -c2 $ext_addr &>/dev/null
+    sudo timeout 4s tcpdump -i $ext_ctrl -w $TMPDIR/ping_switch.pcap &>/dev/null &
+    TIMEOUT_PID=$! # save the pid while this runs in background
+    sleep 1 # a second for tcpdump to settle
+    sudo arp -d $ext_addr || true
+    ping -c4 $ext_addr &>/dev/null || true
+    tail --pid=$TIMEOUT_PID -f /dev/null # wait for timeout+tcpdump to finish writing to pcap file
 fi
 
-# tar and compress, then delete the working folder
-tar -cjf $TMPDIR.tar.bz2 $TMPDIR
+
+# Remove GCP credentials from what is being collected
+if [ -n "$gcp_cred" ]; then
+    rm -f $TMPDIR/$gcp_cred || true
+fi
+
+# Tar and compress the tmpdir, and then delete it
+tar -C $(dirname $TMPDIR) -cjf techsupport.tar.bz2 $(basename $TMPDIR)
 rm -rf $TMPDIR
-echo Created tech-support file [$TMPDIR.tar.bz2]
+echo Created $ROOT/techsupport.tar.bz2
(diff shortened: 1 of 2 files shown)


bin/techsupport
```diff
@@ -0,0 +1,35 @@
...
+fi
+
+if [ -f local/system.conf ]; then
+    source local/system.conf
+fi
+
+# If an external switch was set up, try pinging it and saving that capture
+if [ -n "$ext_ctrl" ] && [ -n "$ext_addr" ]; then
+    echo Saving packet capture of ping to external switch
+    sudo timeout 3s tcpdump -i $ext_ctrl -w $TMPDIR/ping_switch.pcap &>/dev/null &
```
grafnu, line 26: might increase to 4s to avoid race conditions on startup -- rather capture a bit more than too little

bin/techsupport
```diff
@@ -0,0 +1,35 @@
...
+sudo ovs-vsctl show > $TMPDIR/ovs_vsctl_show.txt
+lsb_release -a > $TMPDIR/lsb_release.txt 2>/dev/null
+ip route > $TMPDIR/ip_route.txt
+ip link show > $TMPDIR/ip_link_show.txt
+arp -n > $TMPDIR/arp.txt
+
+if [ -d inst ]; then
+    echo Collecting run logs and config from 'inst' and 'local'
+    cp -r inst $TMPDIR
+    cp -r local $TMPDIR
```
henry54809, line 16: There may be gcp credentials in local which should be omitted?

bin/techsupport
```diff
@@ -0,0 +1,35 @@
...
+    sudo timeout 3s tcpdump -i $ext_ctrl -w $TMPDIR/ping_switch.pcap &>/dev/null &
+    sleep 1 # give a second for tcpdump to settle
+    sudo arp -d $ext_addr
+    ping -c2 $ext_addr &>/dev/null
+fi
+
+# tar and compress, then delete the working folder
+tar -cjf $TMPDIR.tar.bz2 $TMPDIR
+rm -rf $TMPDIR
+echo Created tech-support file [$TMPDIR.tar.bz2]
```
grafnu, line 35: Usually filenames are just there, not sure we need the enclosing []   (generally that's used for something considered optional)

bin/techsupport
```diff
@@ -0,0 +1,35 @@
...
+mkdir -p $TMPDIR
+
+echo Collecting system information
+sudo ovs-vsctl show > $TMPDIR/ovs_vsctl_show.txt
+lsb_release -a > $TMPDIR/lsb_release.txt 2>/dev/null
+ip route > $TMPDIR/ip_route.txt
+ip link show > $TMPDIR/ip_link_show.txt
+arp -n > $TMPDIR/arp.txt
+
+if [ -d inst ]; then
```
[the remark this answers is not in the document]
grafnu: Not sure why I said that... it's already doing the needful.

bin/techsupport
pbatta: good point, omitting them now

bin/techsupport
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash -e
+
+TMPDIR=./techsupport
```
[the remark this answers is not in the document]
pbatta: thanks, addressed.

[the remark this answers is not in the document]
pbatta: moved the temporary directory to /tmp and added techsupport.tar.bz2 to .gitignore

bin/techsupport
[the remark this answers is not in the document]
pbatta: fixed, thanks.

bin/techsupport
pbatta: addressed.

bin/techsupport
pbatta: fixed, thanks.

bin/techsupport
```diff
@@ -0,0 +1,60 @@
...
+save_to_file "ip route"
+save_to_file "ip link show"
+save_to_file "arp -n"
+save_to_file "git describe --dirty"
+save_to_file "git status"
+save_to_file "git remote -v"
+save_to_file "git log -n5"
+
+if [ -d inst ]; then
+    echo Collecting runtime info from 'inst'
```
grafnu, line 27: quotes will have no effect

bin/techsupport
```diff
@@ -0,0 +1,60 @@
...
+    cp -r inst $TMPDIR
+fi
+
+if [ -d local ]; then
+    echo Collecting config from 'local'
```
grafnu, line 32: quotes will have no effect

bin/techsupport
```diff
@@ -0,0 +1,60 @@
...
+
+# If an external switch was set up, try pinging it and saving that capture
+if [ -n "$ext_ctrl" ] && [ -n "$ext_addr" ]; then
+    echo Saving packet capture of ping to external switch
+    sudo timeout 4s tcpdump -i $ext_ctrl -w $TMPDIR/ping_switch.pcap &>/dev/null &
+    TIMEOUT_PID=$! # save the pid while this runs in background
+    sleep 1 # a second for tcpdump to settle
+    sudo arp -d $ext_addr || true
+    ping -c4 $ext_addr &>/dev/null || true
+    tail --pid=$TIMEOUT_PID -f /dev/null # wait for timeout+tcpdump to finish writing to pcap file
```
grafnu, line 48: you can just use "wait" here... (bash command)

bin/techsupport
```diff
@@ -0,0 +1,60 @@
...
+fi
+
+
+# Remove GCP credentials from what is being collected
+if [ -n "$gcp_cred" ]; then
+    rm -f $TMPDIR/$gcp_cred || true
```
grafnu, line 54: this will be insufficient, because there can be other creds in there:

~/daq$ fgrep -l \"private_key\" local/*.json
local/bacnet-gateway-daq-testlab.json
local/bacnet-gateway-peringknife-test.json
local/bacnet-gateway-registrar-test.json
local/bos-daq-testing.json
local/daq-qualification-labs.json
local/essential-keep-197822.json

bin/techsupport
pbatta: fixed, thanks

bin/techsupport
pbatta: fixed, now clearing private_key (and private_key_id) lines from all the json files in local

grafnu: I just remembered there are also keys in inst/  -- so might need to scrape that folder as well.

bin/techsupport
```diff
@@ -0,0 +1,55 @@
...
+fi
+
+if [ -f local/system.conf ]; then
+    source local/system.conf
+fi
+
+# If an external switch was set up, try pinging it and saving that capture
+if [ -n "$ext_ctrl" ] && [ -n "$ext_addr" ]; then
+    echo Saving packet capture of ping to external switch
+    sudo timeout 4s tcpdump -i $ext_ctrl -w $TMPDIR/ping_switch.pcap &>/dev/null &
```
grafnu, line 45: this needs to be n+1 where n is the number of pings. Ping takes ~(n-1) time, but then you add one. So, right now you're still on the harry edge of the race condition.

bin/techsupport
grafnu: I remembered that there's also cretendials places in inst/

bin/techsupport
pbatta: added a check. thanks.

bin/techsupport
pbatta: updated. 5 seconds of tcpdump, 1sec of sleep, 3 sec of pings, so we are at 1 second of wiggle room now

PR #408, opened 2020-05-07
Draft for MUD test changes

didovesei:


testing/test_base.sh
```diff
@@ -50,35 +50,38 @@ echo switch ping $count | tee -a $TEST_RESULTS
...
-device_xcast="$device_traffic and ether dst 9a:02:57:1e:8f:03"
-cntrlr_traffic="tcpdump -en -r inst/run-port-02/scans/monitor.pcap port 47808"
-cntrlr_bcast="$cntrlr_traffic and ether broadcast"
-cntrlr_ucast="$cntrlr_traffic and ether dst 9a:02:57:1e:8f:01"
-cntrlr_xcast="$cntrlr_traffic and ether dst 9a:02:57:1e:8f:03"
+port_1_traffic="tcpdump -en -r inst/run-port-01/scans/monitor.pcap port 47808"
+port_1_bcast_2="$port_1_traffic and ether src 9a:02:57:1e:8f:02 and ether broadcast"
+port_1_ucast_2="$port_1_traffic and ether src 9a:02:57:1e:8f:02 and ether dst 9a:02:57:1e:8f:01"
+port_1_bcast_3="$port_1_traffic and ether src 9a:02:57:1e:8f:03 and ether broadcast"
+port_1_ucast_3="$port_1_traffic and ether src 9a:02:57:1e:8f:03 and ether dst 9a:02:57:1e:8f:01"
```
grafnu, line 57: I'm wondering if we should be pedantic here and also check for a ucast that's NOT destined for :01 -- e.g. make sure that we don't receive unintended packets?  so a port_1_ncast_2 (and _3) that has a not ether dst filter on it... ?

testing/test_base.sh
```diff
@@ -50,35 +50,38 @@ echo switch ping $count | tee -a $TEST_RESULTS
...
-    echo device $type $(($bcast > 2)) $(($ucast > 2)) $(($xcast > 0)) | tee -a $TEST_RESULTS
-    bcast=$($cntrlr_bcast | wc -l)
-    ucast=$($cntrlr_ucast | wc -l)
-    xcast=$($cntrlr_xcast | wc -l)
-    echo cntrlr $type $(($bcast > 2)) $(($ucast > 2)) $(($xcast > 0)) | tee -a $TEST_RESULTS
+    bcast_2=$($port_1_bcast_2 | wc -l)
+    ucast_2=$($port_1_ucast_2 | wc -l)
+    bcast_3=$($port_1_bcast_3 | wc -l)
+    ucast_3=$($port_1_ucast_3 | wc -l)
+    echo port_1 $type $(($bcast_2 > 2)) $(($ucast_2 > 0)) $(($bcast_3 > 2)) $(($ucast_3 > 0)) | tee -a $TEST_RESULTS
```
grafnu, line 73: I'm wondering if these all should be >0 -- not sure why they were >2 initially. What's the intent here?

testing/test_base.sh
```diff
@@ -50,35 +50,38 @@ echo switch ping $count | tee -a $TEST_RESULTS
...
+    ucast_3=$($port_2_ucast_3 | wc -l)
+    echo port_2 $type $(($bcast_2 > 2)) $(($ucast_2 > 0)) $(($bcast_3 > 2)) $(($ucast_3 > 0)) | tee -a $TEST_RESULTS
     more inst/run-port-*/nodes/*/activate.log | cat
 }
 
 test_mud open
 test_mud base
-test_mud todev
-test_mud frdev
+test_mud frdev-mix
```
grafnu, line 84: I'd keep a simple name for now... not sure what additional information the conjunction is adding. Maybe just "mixed"?

grafnu, line 84: Should there also be some removed descriptor files since todev is no longer used?

testing/test_base.sh
didovesei: I think with a specific dst MAC filter it can filter out the broadcast traffic. Or what we can do is something like "ether src xxxx:01 and **not** ether broadcast"?

didovesei: I think specifying the destination will also help filter out the flooding packet for unlearned destination.

testing/test_base.sh
didovesei: I just copied it from the origin test... I thought it was something related to the flooding for learned destination.. But this explanation does not seem to make sense... Changed to 0.

testing/test_base.sh
didovesei: Changed to mixed.

didovesei: Yes, I was not sure if this is what we want to do... Removed.

grafnu:
Closing as stale for now.  Can reopen/restart as necessary.

PR #409, opened 2020-05-08
script for setting up / resetting service account permissions

henry54809:


bin/setup_service_account_permissions
```diff
@@ -0,0 +1,48 @@
...
+   echo Project is not defined.
+   usage
+fi
+
+if [ -z $service_account ]; then
+   echo Service account is not defined.
+   usage
+fi
+
+roles=$(gcloud projects get-iam-policy daq-haoli --flatten="bindings[].members" --format='table(bindings.role)' --filter="bindings.members:$service_account" | tail -n +2)
```
grafnu, line 20: daq-haoli?

bin/setup_service_account_permissions
```diff
@@ -0,0 +1,48 @@
...
+while read -r role; do
+   if [ -z "$role" ]; then continue; fi
+   echo "Removing existing role $role"
+   gcloud projects remove-iam-policy-binding $project --member=serviceAccount:$service_account --role=$role > /dev/null
+done <<< "$roles"
```
grafnu, line 25: can you move the "$roldes" to before the while -- I think it's just "echo $roles | while ..." -- reason is it's a lot more straightforward in terms of flow... the trailing "$roles" is very counterintuiative!

Or maybe it works with: while read -r role <<<"$roles"; do ?

bin/setup_service_account_permissions
```diff
@@ -0,0 +1,48 @@
...
+fi
+
+if [ -z $service_account ]; then
+   echo Service account is not defined.
+   usage
+fi
+
+roles=$(gcloud projects get-iam-policy daq-haoli --flatten="bindings[].members" --format='table(bindings.role)' --filter="bindings.members:$service_account" | tail -n +2)
+while read -r role; do
+   if [ -z "$role" ]; then continue; fi
```
grafnu, line 22: break if onto separate lines

bin/setup_service_account_permissions
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash -e
+project=$1
+service_account=$2
+custom_role_name="CustomStorageRole"
+function usage {
+   echo Usage: setup_service_account_permissions project service_account
```
grafnu, line 6: maybe just set_roles?  The long form is a bit unwieldy (and it will get in the way of directory listings, etc...)

bin/setup_service_account_permissions
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash -e
+project=$1
+service_account=$2
+custom_role_name="CustomStorageRole"
```
grafnu, line 4: DaqStorageRole?

henry54809:
PTAL

PR #418, opened 2020-05-13
Introduce YAML config file

grafnu:


misc/system.yaml
```diff
@@ -0,0 +1,27 @@
+---
+# Base configuration file for DAQ.
+#
+# This setup will create and test one faux device. To specify a real USB
+# adapter, modify the intf_names entry with the appripriate interface name(s).
+# The startup_cmds line can then be commented out to prevent the (unnecessary)
+# startup of the faux test device.
+
+# Description description for dashboard.
```
henry54809, line 9: this duplicated description is still there

proto/system_config.proto
```diff
@@ -0,0 +1,108 @@
+/*
+ * Specification for DAQ configuration.
+ */
+syntax = "proto3";
+
+/*
+ * DAQ configuration.
+ */
+message ForchConfig {
+  // Description description for dashboard.
```
henry54809, line 10: same  here.

proto/system_config.proto
```diff
@@ -0,0 +1,108 @@
...
+  // Location of mud files to use for this setup.
+  string mud_files = 26;
+
+  // File to use for runtime device behavior specification.
+  string device_specs = 27;
+
+  // Configuration directory for runtime tests.
+  string test_config = 28;
+
+  // Set port-debounce for flaky connecitons. Zero to disable.
```
henry54809, line 94: typo

misc/system.yaml
grafnu: updated

proto/system_config.proto
grafnu: updated

proto/system_config.proto
grafnu: fixed

grafnu:
Updated PTAL

PR #421, opened 2020-05-14
leave venv for mininet build 

henry54809:
python2 setuptools is installed outside of venv

### DIFF
--- a/bin/setup_dev
+++ b/bin/setup_dev
@@ -79,6 +79,33 @@ $AG install \
 $AG install python-pip
 python2 -m pip install setuptools
 
+if [ -d mininet ]; then
+    echo Checking mininet version matches $MININETV...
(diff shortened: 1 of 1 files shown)


bin/setup_dev
```diff
@@ -113,7 +113,7 @@ $PIP install --upgrade --index-url=https://pypi.python.org/simple Jinja2 \
     google-cloud-logging==1.14.0
 
 $PIP freeze
-
+deactivate
```
grafnu, line 116: can this be moved inside the subshell for the build, e.g. around line 133?  I'd be worried that this will cause problems later down the line since this deactivation is persistent and extends past the mininet build

henry54809: I think it's just easier to move mininet installation prior to venv activation.

PR #423, opened 2020-05-15
Feature/allow port toggle

henry54809:


daq/docker_test.py
```diff
@@ -100,15 +100,12 @@ def start(self, port, params, callback, finish_hook):
             raise e
         LOGGER.info("Target port %d test %s running", self.target_port, self.test_name)
 
-    def terminate(self, expected=True):
+    def terminate(self):
```
grafnu, line 103: Why is expected going away?

daq/runner.py
```diff
@@ -64,6 +67,7 @@ def __init__(self, config):
         self.exception = None
         self.run_count = 0
         self.run_limit = int(config.get('run_limit', 0))
+        self._default_port_toggle_timeout_sec = int(config.get('port_toggle_timeout_sec', 0))
```
grafnu, line 70: not sure "toggle" is the right word... this is for when the port is explicitly brought down, and then back up again, right?  Network term for that is to "flap" or "bounce" the port... port_flap_timeout_sec?

daq/runner.py
```diff
@@ -262,11 +269,22 @@ def _handle_system_idle(self):
         if all_idle:
             LOGGER.debug('No active device ports, waiting for trigger event...')
 
+    def _monitor_disconnected_port(self):
```
grafnu, line 272: _reap_stale_ports()?

daq/docker_test.py
henry54809: The docker container status can be unknown which is why it’s best for docker_test to figure out if there is still a container to terminate. There’s an exception that’s thrown every time a host is shutdown caused by external factors like timeout. This fixes that problem.

grafnu: How is this case (ip flap timeout) fundamentally different than the module timeout case? At this level, it's a deterministic system so "can be unknown" is not a good sign. And "best" is subjective (and something I disagree with); certainly it's easier, but for a robust system that's often not the right thing. Before altering this, I'd want to know why/how this is fundamentally different than the other cases that happen, which don't put the system in an unknown state.

daq/runner.py
```diff
@@ -299,12 +317,14 @@ def main_loop(self):
...
             self.stream_monitor = monitor
-            self.monitor_stream('faucet', self.faucet_events.sock, self._handle_faucet_events)
+            self.monitor_stream('faucet', self.faucet_events.sock, self._handle_faucet_events,
+                                priority=10)
             if self.event_trigger:
                 self._flush_faucet_events()
             LOGGER.info('Entering main event loop.')
             LOGGER.info('See docs/troubleshooting.md if this blocks for more than a few minutes.')
             while self.stream_monitor.event_loop():
+                self._monitor_disconnected_port()
```
grafnu, line 327: Seems like this can/should be pushed down into the _module_heartbeat() function/system?  Why does this need to be its own top-level thing.  Seems like this (ip timeout) is a related case to the overall module timeout case... would like to keep them as similar as possible.

daq/runner.py
```diff
@@ -35,12 +35,15 @@ class owns the main event loop and shards out work to subclasses."""
...
+        # pylint: disable=too-many-statements
         self.config = config
         self.port_targets = {}
         self.port_gateways = {}
         self.mac_targets = {}
         self.result_sets = {}
-        self._active_ports = {}
+        self._active_ports = set()
+        self._mac_port_mapping = {}
+        self._toggling_ports = {}
```
grafnu, line 46: I think we should move away form multiple top-level things that have the same key.  E.g., I think just one map (not set) for _active_ports would work, and then the "toggling ports" timeout information could be a field inside of that.  In the long run, it makes management of state a lot easier because when you, e.g. delete an active port, you don't have to manage the state in so many different places.

daq/runner.py
```diff
@@ -189,32 +193,35 @@ def _handle_port_state(self, dpid, port, active):
...
+            self._active_ports.remove(port)
 
     def _direct_port_traffic(self, mac, port, target):
         self.network.direct_port_traffic(mac, port, target)
 
     def _handle_port_learn(self, dpid, port, target_mac):
         if self.network.is_device_port(dpid, port):
             LOGGER.info('Port %s dpid %s learned %s', port, dpid, target_mac)
-            self._activate_port(port, target_mac)
+            self._mac_port_mapping[port] = target_mac
```
grafnu, line 224: variable name is backwards, it should be _port_mac_mapping (also, see previous comment about combining this with active ports). It might be that the top-level construct should be "port_info" not "active ports", and "is_active" would just be a sub-entry in the top-level dictionary.

testing/test_aux.sh
```diff
@@ -191,4 +195,65 @@ head inst/run-port-*/finish/nmap*/*
...
+            eval $1 
+            break
+        fi
+        test_done=$(cat $TEST_RESULTS | grep "Done with tests")
+        if [ -n "$test_done" ]; then
+            break
+        fi
+    done &
+}
+rm -f inst/cmdrun.log
```
grafnu, line 213: probably want to put this inside of monitor_interface so it gets cleared each time (otherwise there might be cruft between invocations)

testing/test_aux.sh
```diff
@@ -191,4 +195,65 @@ head inst/run-port-*/finish/nmap*/*
...
+# Check port toggling does not cause a shutdown 
+cp misc/system_base.conf local/system.conf
+echo "port_toggle_timeout_sec=10" >> local/system.conf
+echo "port_debounce_sec=0" >> local/system.conf
+monitor_interface "sudo ifconfig faux down;sleep 1; sudo ifconfig faux up" "Port 1 dpid 2 is now active"
+cmd/run -s
```
grafnu, line 219: Is there a way to combine these 3x cases into one + unit_test coverage?  Seems like there's only really two behavioral cases that need to be captured -- flap is fast (no problem), or flap is too slow (errors out) -- everything else (different settings of values) should be accomplished with a unit test (so no need to run cmd/run).  Anurag added the unit test framework in, so it should be easier to use that.

I'm trying to stem the testing bloat that will come from too many integration tests for everything. They're "easy" -- but also really slow and add a lot of technical overhead to the system in the long run.

testing/test_aux.sh
```diff
@@ -191,4 +195,65 @@ head inst/run-port-*/finish/nmap*/*
 
 tcpdump -en -r inst/run-port-01/scans/test_nmap.pcap icmp or arp
 
+
+function monitor_interface {
```
grafnu, line 199: THis doesn't seem to be "monitor_inteface" -- it's really "monitor_log"?  The interface part is only incidental b/c of how it's called, not the function itself.

daq/docker_test.py
henry54809: The fundamental difference is that module timeout kills the test, and port flap timeout kills the host and all the remaining tests. When killing the host, which happens at the runner level, runner currently always assumes all test containers are terminated under 

> target_set_cancel -> target_host.terminate(... trigger=False)-> self.test_host.terminate(expected=False) -> if bool(self.docker_host) != expected: raise exception 

It is usually not the case during a port flap that a container is already terminated so that assumption causes an exception to be raised. To fix that issue, I considered moving that logic to host; however even at the host level, the container running status can be unknown. 
Host only knows the container termination after the docker_test callback; however, there is a lag between the container is terminated (stream hangup) and the callback is called. Depending on the order these events are processed, host may not know a running container is already terminated when it tries to terminate the container again.

daq/runner.py
henry54809: Module time out should be at the host level and 1 test timing out should have no effect on the remaining tests. Port flap timeout should be at the runner level since the timeout for that should result in the killing of the host.

grafnu:
I would suggest creating a separate doc with a (simple) bulleted list of
the sequence of events. You're describing a fair bit of nuance there, which
could very well be accurate, but I don't get the assertion of "container
running status can be unknown" -- why can't it be known?  Yes, events can
be processed in different order, but depending on which one gets their
first then it should be able to resolve things cleanly. Maybe I'm missing
something, which is why I'm suggesting calling it out in a separate doc
(comments on a PR aren't the best way to diagram/discuss sequences of
things).   The state seems like it should be known somewhere in the system
because something is clearing out self.docker_host, right?  Is it a
matter/question then of how that state change isn't being propagated to
other parts of the system in a deterministic fashion?

grafnu:
Got it, makes sense. (Timeout at runner level for this.)

henry54809:
PTAL.

henry54809:
@grafnu everything’s addressed in this commit.

PR #429, opened 2020-05-19
Add password switch config

jhughesbiot:


jhughesbiot:
add preliminary password switch config

PR #433, opened 2020-05-20
Firebase db Docs update

henry54809:


docs/firebase.md
```diff
@@ -230,3 +230,6 @@ DAQ -> Firebase: Merged (active)
 Firebase -> WebApp: Merged (active)
 note right of WebApp: "Normal" state
 ```
+
+# Firebase Database Schema
+![Diagram](Firebase%20Database.png)
```
grafnu, line 235: What's the source representation for this image?  I'm hesitant to use imported drawings for anything of substance because it's not really "source" and makes it harder to maintain in the long run. Really would like to have a textual representation of whatever it is so it can be managed w/ version control...

henry54809:
PTAL

docs/firebase_schema
```diff
@@ -0,0 +1,76 @@
+Table User as U {
+  id uuid [pk]
+  email string
+  name string
+  updated timestamp
+}
+
+Table Permission as P{
```
grafnu, line 8: nit: missing space after P

docs/firebase.md
```diff
@@ -230,3 +230,7 @@ DAQ -> Firebase: Merged (active)
...
+Generated with [dbdiagram.io](https://dbdiagram.io/) with this [file](firebase_schema)
```
grafnu, line 236: can you add an extension to the firebase_schema?  I don't care that it's the right one, necessarily -- just don't want it to be blank, wrong, or misleading.

PR #437, opened 2020-05-21
update techsupport to use new switchconfig names. include date-time in name of created file

pbatta:


daq/runner.py
```diff
@@ -273,7 +273,7 @@ def _handle_system_idle(self):
             LOGGER.debug('No active device ports, waiting for trigger event...')
 
     def _reap_stale_ports(self):
-        for port, port_info in self._port_info.items():
```
grafnu, line 276: did this get pulled in from another change?

pbatta: It is unrelated to the tech-support changes, but came up as an integration test failure for this PR. Shall I split it out to a separate PR?

henry54809: That was for fixing the “dict size changed during iteration problem” in aux test.

grafnu:
The question was not "what is this change" but rather "why is it in this
PR" -- and was it *intentional* or inadvertent -- @puneet, if you're making
this change intentionally then I think it's OK since it's small.
Theoretically things should be separated out into indivisible chunks, but
that does have an overhead for it, so no hard and fast rule for something
this small.

PR #447, opened 2020-05-27
Convert PortInfo dict to class

grafnu:


daq/runner.py
```diff
@@ -97,14 +106,11 @@ def _get_states(self):
...
         return states + connected_host.post_states()
 
-    def _get_active_ports(self):
-        return list(filter(lambda p: self._port_info[p]["active"], self._port_info.keys()))
-
     def _send_heartbeat(self):
         message = {
             'name': 'status',
             'states': self._get_states(),
-            'ports': self._get_active_ports(),
```
henry54809, line 107: This is meant to send port activity information to the UI. Sending just the running ports will break that functionality.

PR #448, opened 2020-05-27
Fix unit tests

anurag6:


testing/run_unit_tests.sh
```diff
@@ -6,13 +6,15 @@ TESTDIR=`dirname $SCRIPTPATH`
...
 
-PYTHONPATH=$BASEDIR/daq:$BASEDIR/mininet:$BASEDIR/faucet:$BASEDIR/forch $BASEDIR/venv/bin/coverage run \
+coverage erase
+
+PYTHONPATH=$BASEDIR/daq:$BASEDIR/mininet:$BASEDIR/faucet:$BASEDIR/forch coverage run \
     --source $BASEDIR/daq \
     -m unittest discover \
     -s $TESTDIR/unit/ \
-    -p "test_*.py"
+    -p "test_*.py" || exit 1
```
grafnu, line 17: exit 1 is redundant with -e option at top

testing/run_unit_tests.sh
```diff
@@ -6,13 +6,15 @@ TESTDIR=`dirname $SCRIPTPATH`
 BASEDIR=`readlink -f $TESTDIR/..`
 cd $BASEDIR
 
-$BASEDIR/venv/bin/coverage erase
+source venv/bin/activate
 
-PYTHONPATH=$BASEDIR/daq:$BASEDIR/mininet:$BASEDIR/faucet:$BASEDIR/forch $BASEDIR/venv/bin/coverage run \
+coverage erase
+
+PYTHONPATH=$BASEDIR/daq:$BASEDIR/mininet:$BASEDIR/faucet:$BASEDIR/forch coverage run \
```
grafnu, line 13: set separately and export -- this is getting ugly as a prefix

testing/run_unit_tests.sh
anurag6: Done.

testing/run_unit_tests.sh
anurag6: Done

PR #459, opened 2020-06-01
Use module_config for module switch configuration

grafnu:


daq/host.py
```diff
@@ -559,37 +558,52 @@ def _docker_test(self, test_name):
...
+            self._record_result(self.test_name, config=self._loaded_config, state=MODE.CONF)
             self.record_result(test_name, state=MODE.EXEC)
             self._monitor_scan(os.path.join(self.scan_base, 'test_%s.pcap' % test_name))
             self.test_host.start(self.test_port, params, self._docker_callback, self._finish_hook)
-        except:
-            self.test_host = None
-            raise
+        except Exception as e:
+            if self.test_host:
+                self.test_host.terminate(optional=True)
```
henry54809, line 592: Is the optional necessary here? From what I can recall, the self.test_host is a direct indicator of whether  the docker_host has been terminated.

grafnu: I think you are half right. test_host is not a direct indicator because the docker_host doesn't exist until the test_host has been started, so during initialization there's a state where the test_host exists but docker_host does not; however, that doesn't mean "optional" was the best fix.  I've refactored the code to break it up into smaller functions, take out the optional, and catch errors at appropriate spots so it can more directly cleanup state.

grafnu:
Updated, PTAL.

PR #461, opened 2020-06-02
security.admin.password changes

em-redstone:
Hi @grafnu, @pisuke, @noursaidi, PTAL at the recent password test changes.

Changelog:
- Ported from java to bash
- Removed macoui dependency, more comprehensive dictionary
- Updated documentation and test description
- Added skip case for hydra errors
- Added experimental web form hydra command script

Many thanks, much appreciated!


grafnu:
This seems to break all the TLS tests (and with committed results)... why?

em-redstone:
Hi Trevor, it seems that the tls tests were failing due to an entry added to the subset/security/module_manifest.json file. The entry has been removed for now, and the tls test works again.

docs/device_report.md
```diff
@@ -407,90 +407,78 @@ RESULT skip security.tls.v1_3.x509 IOException unable to connect to server
...
-Starting NMAP check...
 
 Starting Nmap 7.60 ( https://nmap.org ) at XXX
 Nmap scan report for daq-faux-1 (X.X.X.X)
 Host is up (XXX).
 
-PORT      STATE SERVICE
-10000/tcp open  snet-sensor-mgmt
+PORT   STATE  SERVICE
+80/tcp closed http
```
grafnu, line 418: Is this change set meant to replace the previous brute-force testing?   I think it is, and if so, I'd like to find another place to test the "oh, and port 10000 was open" -- this could be simply done in the testing/test_modules.sh file by adding another line (or option to existing line) for the nmap line that invokes some code in docker/include/bin/start_faux that replaces the existing 'brute' and 'nobrute" keywords.

subset/security/password/test_password
```diff
@@ -0,0 +1,81 @@
+#!/bin/bash -e
```
grafnu, line 1: How is this different than the subset/pentests/test_brute code?  I'm just trying to make sure we're not ending up with duplicate stuff -- should test_brute be removed?

docs/device_report.md
em-redstone: After adding some changes to the java code in the previous merge of the password test, I found that all it did was really call third party brute force tools such as ncrack and hydra. I think that Java may be a bit overkill for this so I ported the code into bash - it still does the same thing as the java code, just hopefully the code is much easier to maintain now. Please note this isn't meant to replace test_brute which I believe was just for educational purposes.

As for the nmap stuff, the password test only runs nmap commands to check if a port is up and if it is, it will run the password test. I think i could probably remove this as I added a skip case for hydra errors now - will have a look.

Thanks, hope this helps.

grafnu:
Ok -- something other than the Java -> Bash conversion is going on, since
there was a test for "find and fail if there was an unexpected open port
(tested with port 10000) -- just wondering where that functionality went!
I want to make sure some part of the system (maybe it's in the nmap scan, I
don't know) is checking for random unexpected open ports.

test_brute shouldn't be "just for educational  purposes" -- so either we
should make sure it's being tested properly, or remove it. Reason is that
having "dead code" is dangerous because it can easily get out of date and
so causes problems (would educate people incorrectly!).

grafnu:
Ok, rather than implicitly checking if "it's the faux device or not" --
there should be an explicit configuration parameter (e.g. which dictionary
to use) that can then be set for testing.
See resources/setups/baseline/module_config.json as an example of how to do
this (for the ip_addr test)... if you set a parameter like that, it will be
available to the docker container module_config.json file (see other
examples). That particular file isn't the right place to set it, but
something in the testing/ suite should do it when there's the particular
test pattern that it hits when it's too slow.  Let me know if you'd like
any help sorting that part out.

em-redstone:
_"there was a test for "find and fail if there was an unexpected open port (tested with port 10000) -- just wondering where that functionality went!"_

I think the reason why the line: "10000/tcp open  snet-sensor-mgmt" disappeared from the device report is because the Java version of the password test did an unrestrained nmap scan on all ports. The bash version now uses nmap to only look for ports 80, 443, 22 and 23 with nmap -p which explains why its no longer there. Its probably still there as the output for the actual nmap test is unchanged.

_"test_brute shouldn't be "just for educational  purposes" -- so either we should make sure it's being tested properly, or remove it."_

My stance on test_brute is that I think it is useful to have some form of educational test in daq, but there are many tests in daq already which can be used as reference and at the same time be part of the spec. Comparing test_brute with test_password, they are both quite similar, except the brute force test does not use third party tools, so some redundancy is introduced. I think if test_brute was removed, it can easily be compensated by looking at test_password instead, or some sort of documentation on how to set up a simple daq test.

em-redstone:
_"The system needs to test somewhere than an arbitrary port being open (e.g.
10000) is discovered and marked as a failure. In the brave-new-world of
bash-rather-than Java, where/when/how is this device behavior (arbitrary
open port) being tested?"_

I see, thanks for clarifying - I don't think even the Java version checked for an arbitrary port, what I remember is that the password test only uses nmap to check if http/https/telnet/ssh ports are open, as when it used ncrack on closed ports it would just hang there.. now that nmap is used it can detect closed ports and return a skip case. I don't think the bash version does checks for arbitrary ports either.

If this is a required feature, I can always add it onto the password test..

testing/test_aux.out
```diff
@@ -10,11 +10,11 @@ RESULT info protocol.bacnet.version Protocol version: 1
 RESULT pass protocol.bacnet.pic The devices matches the PICS
 RESULT fail connection.mac_oui Manufacturer prefix not found!
 RESULT pass connection.mac_oui Manufacturer: Google found for address 3c:5a:b4:1e:8f:0a
-RESULT skip security.tls.v1 IOException unable to connect to server
+RESULT skip security.tls.v1 IOException unable to connect to server
```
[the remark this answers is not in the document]
em-redstone: I've run grep on the daq folder to find that the line of code which outputs this, "daq/subset/security/tlstest/src/main/java/Certs.java" (this may also address the other comment about the Host: xxxx line by looking at the last few commits on this Certs.java) was already outputting a space at the end even before this pull request. I don't think the code in this pull request modifies any of the tls tests.

It could just be that test_aux.out has not been updated in a while and incremental changes were added during travis testing for it to pass.

Hope this helps.

grafnu:
Ok -- I took a closer look at everything and found a few problems with
things that were causing this.  The difficult part is that it causes
unexpected changes in your PR that make it hard to really see what's going
on -- e.g. I end up asking lots of questions about "why" something is there
when it really has nothing to do with your stuff.  I put together
https://github.com/faucetsdn/daq/pull/468 -- once that's in, it should
clear things up.

subset/security/password/test_password
pisuke: Yes, test_brute should be removed as it is not doing anything useful. We kept it so far only for "educational" reasons.

pisuke:
> Ok -- something other than the Java -> Bash conversion is going on, since there was a test for "find and fail if there was an unexpected open port (tested with port 10000) -- just wondering where that functionality went! I want to make sure some part of the system (maybe it's in the nmap scan, I don't know) is checking for random unexpected open ports.

Yes, the nmap scan test is the right place for doing this.

pisuke:
> test_brute shouldn't be "just for educational purposes" -- so either we should make sure it's being tested properly, or remove it. Reason is that having "dead code" is dangerous because it can easily get out of date and so causes problems (would educate people incorrectly!).

I agree we should remove it as it is not functional and not in line with the test plan.

em-redstone:
> Ok, rather than implicitly checking if "it's the faux device or not" -- there should be an explicit configuration parameter (e.g. which dictionary to use) that can then be set for testing. See resources/setups/baseline/module_config.json as an example of how to do this (for the ip_addr test)... if you set a parameter like that, it will be available to the docker container module_config.json file (see other examples). That particular file isn't the right place to set it, but something in the testing/ suite should do it when there's the particular test pattern that it hits when it's too slow. Let me know if you'd like any help sorting that part out.

Hi Trevor, 

I've modified the test_password script to look for the field "use_faux_dictionary" in the container module config and removed the implicit check. It will use the faux dictionaries if that field is set to true, and the full version otherwise. 

To get the configuration into the container, I've modified testing/test_aux.sh to create a copy of the baseline module config in the local folder, modify it to have the "use_faux_dictionary" field and set it as base_conf in system.yaml.

The other changes are related to the password test validation.

Please take a look and let me know if this is the right way to do it,

Many thanks!

grafnu:
Almost!  It shouldn't be a boolean value and it shouldn't be specific for
faux -- it should just be something like "dictionary_file" or whatever, and
then the value would be the necessary path.  If there's multiple files, you
could make it "dictionaries_dir" to point at the directory where all the
actual files would be. So, rather than the current file
"resources/faux_dictionary.txt" it would be something more like
"resources/faux/dictionary.txt" and then point the parameter at the right
thing for "resources/faux" (actual values may vary because of internal path
mappings).  You could then also have resources/full/dictionary.txt as the
default.

The conceptual problem now is you're still bleeding the "faux" abstraction
into the "test_password" subsystem... Nothing inside of the password test
should directly know about faux -- only through indirect things like a
configuration directory.

Everything else looks good!

em-redstone:
> Almost! It shouldn't be a boolean value and it shouldn't be specific for faux -- it should just be something like "dictionary_file" or whatever, and then the value would be the necessary path. If there's multiple files, you could make it "dictionaries_dir" to point at the directory where all the actual files would be. So, rather than the current file "resources/faux_dictionary.txt" it would be something more like "resources/faux/dictionary.txt" and then point the parameter at the right thing for "resources/faux" (actual values may vary because of internal path mappings). You could then also have resources/full/dictionary.txt as the default. The conceptual problem now is you're still bleeding the "faux" abstraction into the "test_password" subsystem... Nothing inside of the password test should directly know about faux -- only through indirect things like a configuration directory. Everything else looks good!

Hi Trevor,

I've changed the script to take in the dictionary folder as a configuration "dictionary_folder". The value is a path to the folder e.g. "resources/faux or resources/default" which have the brute force files. It is also possible to provide a custom port for each protocol now (should there be a need) via the config.

The other changes are just some refactoring and renaming to make the code and file names clearer, as well as added documentation.

Regards.

em-redstone:
Hi Trevor,

Thanks for the comments. I've refactored the code to replace instances of 'folder' with 'dir' now.

I've also noticed I've been getting these extra lines in test_aux.out, even in the master branch:

(Please see the latest PR commit)
https://github.com/faucetsdn/daq/pull/461/commits/16f2fb2d29fac62e0f430b28ad172b53df0a3776

I'm not sure what produces these, but I have just been removing them for now since I see this in master branch runs too. Is this normal, or could it be differences to how DAQ has been set up on my machine?

Many thanks,

Edzel.

grafnu:
Edzel, I'm not 100% sure about the differences... I'm doing a run myself to check it out, then I'll let you know.  Other than that, everything looks good thanks!  Just waiting for the tests to clear then I'll approve.

grafnu:
Looks like it's something about how DAQ is setup on your machine... the
tests as are in the PR are working fine.

em-redstone:
Thanks Trevor, sounds great that it's only a local issue.

Regards

subset/security/password/run_password_test_for_protocol
```diff
@@ -0,0 +1,165 @@
...
+  echo "$1" | grep -q "Host seems down"
+}
+
+# Checks if http or https has authentication.
+# $1 Target protocol (HTTP or HTTPS)
+# $2 Target IP
+# $3 Target port
+function does_http_or_https_have_authentication() {
+  if [[ $1 == "http" || $1 == "https" ]]; then
+    echo "$(curl -k --silent $1://$2:$3)" | grep -q "401 Authorization Required"
```
noursaidi, line 53: @em-redstone  you should look in the headers for 401 response code rather than the HTML body, as the text varies between servers. For example Apache will instead say “401 Unauthorized”, which results in a test skip. You could add the '-i' flag to the curl request

em-redstone: Hi @noursaidi,

I've added the changes - It now only looks in the headers for the response code.

I've also heard mention in the meeting, upcoming features (operator to decide which password file to use) to be added to the password module - Will that be the only feature to be added, or should more be expected to come? Also who will now be expected to implement them?

Many thanks.

noursaidi: Thanks @em-redstone, looks good and it all works now as expected. Thanks for incorporating this comment and all the other comments we've had.

I think the ability for a test operator to provide a username & password combination is the final feature which is needed and completes the test, I can't see any other changes required currently. I'm happy to incorporate this additional functionality

em-redstone: With regards to this feature, the module recently now has the option to specify which files the password test will use in the module_config.json (Please see readme.md for more info) so I actually think it's been implemented already.

Currently there are two sub folders within the resource folder, which is copied into the docker container.
- faux: Which contains a shortened version of the password list for faux device testing. The script testing/test_aux.sh updates the config to use this folder instead of the default.
- default: The full password list currently used by the module.

If the operator needs to add their own - ideally they will just need to make a new folder in there with their own version the files, and update the module_config.json

The files needed would be:
- A .txt file containing the usernames and password pairs in the format :username:password
- Two separate files containing the same usernames and passwords in the above list.

Reason for the multiple files is that ncrack doesn't seem to support colon separated files.

Many thanks.

noursaidi: Thanks @em_redstone, I noticed that change when looking at the code, and is a very useful and welcome if someone wanted a more tailored password list and to deal with the length of the test. The change we had in mind was to expand on it, instead of specifying a file they specify the username and password directly along with the device serial, etc, in module_config

```
“default_username”: “*** (optional) ***”,
“default_password”: “*** (optional) ***”
```

Then, in test_password you could have these into a new dictionary with a single line. I haven't tested this, but it i hope it explains what I had in mind.

```
USERNAME_KEY=".device_info.default_username"
PASSWORD_KEY=".device_info.default_password"

DEFAULT_USERNAME="$(get_module_config_value_from_key $MODULE_CONFIG $USERNAME_KEY)"
DEFAULT_PASSWORD="$(get_module_config_value_from_key $MODULE_CONFIG $PASSWORD_KEY)"

if [[ -z "${$DEFAULT_USERNAME}" ]]
    echo "${DEFAULT_USERNAME}:${DEFAULT_PASSWORD}:" >> provided_credentials.txt
    NEW_DICTIONARY_DIR = provided_credentials.txt
fi 
```

em-redstone: Hi @noursaidi,

Will the other feature mentioned in the spec sheet involving the output of the different protocols to be merged into a single result be required?

Many thanks.

noursaidi: @em-redstone I believe when the tests were renamed recently they were retained as separate tests so I thought not. @pisuke @spectre-storm please confirm?

em-redstone: Hi @noursaidi,

Please take a look at the recent code change.

The module can now use the specified "default_username" and "default_password" value in the module config to use as the only combination when brute forcing.

The process added involves creating another folder called "resource/user_specified" and its constituent files "dictionary.txt", "usernames.txt" and "passwords.txt" with the username/password specified as the only values.

Readme.md for this file has also been updated.

Many thanks.

noursaidi:
Thanks @em-redstone , all looks good to me. Thanks for resolving all the comments. @pisuke do you have any additional comments?

pisuke:
All good for me too, thanks!

PR #479, opened 2020-06-12
convert device_types to device_usage

grafnu:


grafnu:
@pisuke, I'm not sure about the changes in (e.g.) resources/setups/qualification/minimal_device_module_config.json  -- it seems to be an extra field that's not used... but maybe I'm missing something?

PR #480, opened 2020-06-12
Report handling cleanup and refactoring

grafnu:


bin/python/combine_reports_from_date_range.py
```diff
@@ -46,34 +48,48 @@ def _render_results(results):
...
         if (start_str and timestamp < start_str) or (end_str and timestamp > end_str):
             LOGGER.info('Skipping file %s' % json_file)
             continue
         LOGGER.info('Processing file %s' % json_file)
         with open(os.path.join(reports_dir, json_file), 'r') as json_file_handler:
             yield json.loads(json_file_handler.read())
 
-def main(device, start=None, end=None, gcp=None, reports_dir=os.path.join('inst', 'reports')):
+
+def main(device, start=None, end=None, gcp=None, reports_dir=DEFAULT_REPORTS_DIR, count=0):
```
henry54809, line 76: Why is there an explicit count requirement?

bin/python/combine_reports_from_date_range.py
```diff
@@ -46,34 +48,48 @@ def _render_results(results):
...
+        json_reports = gcp.get_reports_from_date_range(device_full, start=start, end=end,
+                                                       count=count)
     else:
-        json_reports = _get_local_reports(device, reports_dir, start, end)
+        json_reports = _get_local_reports(device, reports_dir, start, end, count)
 
+    aggregate = {'tests': {}, 'categories': {}, 'missing': {}}
+    report_count = 0
     for json_report in json_reports:
+        report_count += 1
```
henry54809, line 92: report_count is not really necessary since report_count is len(json_reports)

grafnu: Kinda -- json_reports is actual an iterator so... doesn't like to use it twice.  But, I fixed that with a list() call instead of explicit count. Maybe less efficient but more straightforward.

bin/python/combine_reports_from_date_range.py
grafnu: Partly for testing, where I wanted to make sure there were exactly N results to pull from for consistency. But also I imagine for a real "soak test" we'd likely want to say "the result of the last 10 runs" since otherwise there could be 1, or 100, which would have wildly different meanings...

grafnu:
Updated, PTAL. Only one small change re: length...

bin/python/combine_reports_from_date_range.py
henry54809: well, the results coming from GCP is not sorted so you can have different combined results with each run.

bin/python/combine_reports_from_date_range.py
```diff
@@ -88,12 +104,15 @@ def main(device, start=None, end=None, gcp=None, reports_dir=os.path.join('inst'
...
+    start_stamp = _iso_to_fname(start)
+    end_stamp = _iso_to_fname(end)
+    reports_path = os.path.join(reports_dir, 'combo_%s_%s_%s.md' % (device, start_stamp, end_stamp))
     with open(reports_path, 'w') as report_file:
         report_file.write('# Combined Results\n')
-        report_file.write('Device: %s Start: %s End: %s\n' % (device, start, end))
+        report_file.write('Device: %s\nStart: %s\nEnd: %s\n' % (device, start, end))
         report_file.write(result_str)
     LOGGER.info('Report written to %s' % reports_path)
+    assert not count or count == len(json_reports), 'Did not find expected %d reports' % count
```
henry54809, line 115: I don't really think the assertion is necessary. It's hard to know how many reports you have if you specify the device, start and end.

bin/python/combine_reports_from_date_range.py
henry54809: I forgot I used a generator on purpose here

grafnu:
It only kicks in if you specify a count, and I wanted to make it obvious if
you weren't getting what you asked for.  I'm not sure what you mean by
"it's hard" -- if I run it for an hour, and then say "give me the last
10"... I'd want to know right away if there were less than 10 available.
Also for testing, I ask for 2, and I want to know if it's not getting what
it's expected.

grafnu:
Pushed a fix for that. I still need to test it completely, but the concept
is at least there.

PR #482, opened 2020-06-15
report refactoring

grafnu:
A bit more sprawling than I'd like, but I wanted to add a proper integration test for the gcp variant and I ran into some things to clean up. No real functionality change, just testing and internal cleanup.


daq/gcp.py
```diff
@@ -252,29 +255,32 @@ def register_offenders(self):
...
+        blob = self._bucket.blob(report_blob)
+        return json.loads(str(blob.download_as_string(), 'utf-8'))
+
     def get_reports_from_date_range(self, device: str, start=None, end=None, count=None):
         """Combine test results from reports within a date range"""
         if not self._firestore:
             LOGGER.error('Firestore not initialized.')
             return
         LOGGER.info('Looking for reports...')
-        origins = self._firestore.collection(u'origin').stream()
```
henry54809, line 261: I think we need to keep pulling from origins so that all reports for this device are considered.

daq/report.py
```diff
@@ -149,7 +153,7 @@ def finalize(self):
         for extension in ['.md', '.pdf', '.json']:
             if self._alt_path:
                 shutil.copyfile(self._report_prefix + extension, self._alt_prefix + extension)
-            report_paths.update({'report_' + extension: self._report_prefix + extension})
+            report_paths.update({'report' + extension: self._report_prefix + extension})
```
henry54809, line 156: The front end (main.js under handleFileLinks) uses "_path" in attribute (e.g. report_path) to determine if it needs to generate a link for file download.

grafnu:
Two things, though...
1) Then the behavior is different from the "run locally" option, which only
applies to local devices. If there's a demonstrable need for a combined
report across multiple origins.
2) It doesn't follow the model for everything else done from the origin,
which interacts with the cloud project only for that origin.
3) There's nothing to say that all origins will be in the same cloud
project, so could have surprising behavior.

All and all, I think if there's a demonstrable need for cross-origin
combined reports, then it would be an explicit option or setup, not
default/baseline behavior.

grafnu:
Thanks for the catch... obviously lacking in QA. Updated, PTAL... will do a
round of UI testing on it soon.

henry54809:
I think it's makes sense to pull all available data but we can revisit it another time.

PR #486, opened 2020-06-16
UDMI fittings RFC for LCS

grafnu:


schemas/udmi/pointset.json
```diff
@@ -30,17 +30,23 @@
...
-          {
-            "enum": [
-              "present_value"
-            ]
+      "additionalProperties": false,
+      "properties": {
+        "present_value": {
+          "type": ["number", "string", "boolean"]
+        },
+        "fittings_value": {
```
pisuke, line 38: Why using fittings_value as a name of this item? It looks like you're creating an array of items so why not just call it array or something similar and abstract it?

grafnu:
Array is way too generic, it's a type, but there's nothing then to
distinguish it from other arrays.  I checked, and "fittings" actually is an
appropriate word because it means a small thing attached to a main thing.
I'm looking for something that's not LCS-specific, but makes sense in that
context and also has some differentiation from other things. Thus, "Array"
is too generic (what happens if there's another array of stuff?). In short,
though... any other candidates for a better word?

schemas/udmi/pointset.json
jcbrough: I think as we delve deeper we will find more examples like this where we have complex objects that do not have a simple 'value'. Time Schedules are a collection of start/stop times in a time/date calendar; Lighting control includes 'scenes' which define light levels for different luminaires. In programming they would an object. Maybe that is a term worth considering?

grafnu:
The concept of an "object" could make sense, but I'd like to carefully
consider it in a specific context (this current change is about adding LCS
point data).  If there are some concrete examples of the schedules and
scenes, we could work through that. There are many different ways to skin
that cat.

   - E.g., if it's start/stop values, then you have two points (start_time
   and stop_time) that can get mapped to the collection of fixtures.
   - I doubt all the scene data would be encoded directly in the UDMI
   payload. Either it would be an abstraction down lower in the stack where
   the "scene" would be a point-name that's passed down. Or, the scene would
   be encoded up somewhere in the higher layers of the stack and it would send
   down the individual point values (so the actual scene string is
   irrelevant).

The concept of a "set of point values for a set of scenes" would be encoded
as part of config data, not as telemetry (which is what this message is) --
so that might be something worth considering. At that point, through, I
don't think it would be a "pointset" anymore, but rather a "scene" which
would be a parallel construction, and then we can do whatever it is we
needed to do for that.

TL;DR: I think being able to express a "scene" could be interesting, but
would likely be something new rather than an iteration on what pointset is.

PR #487, opened 2020-06-17
NTPv4 support

frgitdaq:


docker/include/bin/start_faux
```diff
@@ -132,9 +132,12 @@ elif [ -n "${options[bacnetfail]}" ]; then
          FauxDeviceEngine.EntryPoint $local_ip $broadcast_ip "Faux-Device-Fail.json"  &
 fi
 
-if [ -n "${options[ntp_client]}" ]; then
+if [ -n "${options[ntpv4fail]}" ]; then
```
grafnu, line 135: can you name the options something to do with how the behavior is different, not the pass/fail result?  Faux itself doesn't really "know" about what the test is, it just knows what its behavior is.

docker/include/bin/start_faux
```diff
@@ -132,9 +132,12 @@ elif [ -n "${options[bacnetfail]}" ]; then
...
     echo Starting ntp client.
     java -jar NTPClient/build/libs/NTPClient-1.0-SNAPSHOT.jar "time.google.com" "123" "3" &
+elif [ -n "${options[ntpv4pass]}" ]; then
+    echo Starting ntp client.
+    java -jar NTPClient/build/libs/NTPClient-1.0-SNAPSHOT.jar "time.google.com" "123" "3" "4" &
```
grafnu, line 140: quotes are redundant (please remove)

subset/network/NTPClient/src/main/java/Main.java
```diff
@@ -19,6 +20,9 @@ public static void main(String[] args) {
         serverName = args[0];
         PORT = Integer.parseInt(args[1]);
         timerPeriod = Integer.parseInt(args[2]);
+        if (args.length == 4) {
```
grafnu, line 23: DOes this need to be optional?  Seems better just to make it required all the time?

Should check for the bad case (# of args != 4) and print a message.

docker/include/bin/start_faux
```diff
@@ -132,9 +132,12 @@ elif [ -n "${options[bacnetfail]}" ]; then
          FauxDeviceEngine.EntryPoint $local_ip $broadcast_ip "Faux-Device-Fail.json"  &
 fi
 
-if [ -n "${options[ntp_client]}" ]; then
+if [ -n "${options[ntpv4fail]}" ]; then
     echo Starting ntp client.
     java -jar NTPClient/build/libs/NTPClient-1.0-SNAPSHOT.jar "time.google.com" "123" "3" &
+elif [ -n "${options[ntpv4pass]}" ]; then
```
grafnu, line 138: as per other comment, can you make the two cases more similar -- e.g. both should have the same number of arguments. I don't see any reason why anything should be optional here.

docker/include/bin/start_faux
```diff
@@ -132,9 +132,12 @@ elif [ -n "${options[bacnetfail]}" ]; then
...
+    echo Starting ntp client.
```
grafnu, line 139: diagnostic string should not be the same in both cases.

subset/network/test_network
```diff
@@ -2,14 +2,15 @@
...
-python network_tests.py connection.dhcp_long $MONITOR $TARGET_IP
-python network_tests.py connection.min_send $MONITOR $TARGET_IP
-python network_tests.py communication.type.broadcast $MONITOR $TARGET_IP
-python network_tests.py protocol.app_min_send $MONITOR $TARGET_IP $MODULE_CONFIG $EXCLUDES
-python network_tests.py network.ntp.support $MONITOR $TARGET_IP
+#python network_tests.py communication.type.broadcast $MONITOR $TARGET_IP
+#python network_tests.py connection.dhcp_long $MONITOR $TARGET_IP
+#python network_tests.py connection.min_send $MONITOR $TARGET_IP
+python network_tests.py connection.network.ntpv4 $STARTUP $TARGET_IP
+#python network_tests.py protocol.app_min_send $MONITOR $TARGET_IP $MODULE_CONFIG $EXCLUDES
```
pisuke, line 14: @frgitdaq it's best to move the ntp code in a separate ntp module (test_ntp) to avoid collisions with other tests.
Leave the previous test code as it is as it will be addressed in a separate issue and PR.

frgitdaq:
@pisuke the NTP testing has been moved into a separate NTP module. The [latest Travis build](https://travis-ci.com/github/faucetsdn/daq/jobs/357311576) is outputting the markdown report generated by the NTP module. Is there a configuration file where the NTP module should be stated so that it is not included in this output?

subset/network/test_network
frgitdaq: @pisuke The code has now been moved into a separate module.

frgitdaq:
@grafnu @pisuke Just wanted to update, the recommended changes have been implemented and the builds as passing on Travis. If there is anything else you are waiting on, then please let me know?

grafnu:
Mostly looks fine -- just one additional lingering lint error...

frgitdaq:
> Mostly looks fine -- just one additional lingering lint error...

@grafnu @pisuke The build has passed with the removal of the lint error

PR #491, opened 2020-06-19
Use update method signature for gcp credentials

grafnu:
Some library changed somewhere, and the ugly "use a private method" hack is no longer necessary (nor possible)... but now we get to use the proper public function!


daq/gcp.py
```diff
@@ -58,7 +58,7 @@ def __init__(self, config, callback_handler):
         LOGGER.info('Loading gcp credentials from %s', cred_file)
         # Normal execution assumes default credentials.
         # pylint: disable=protected-access
```
henry54809, line 60: Maybe this pylint isnt needed either

PR #494, opened 2020-06-22
Baseline for NTP tests

pbatta:


docker/include/bin/start_faux
```diff
@@ -141,6 +141,26 @@ if [ -n "${options[ntp_client]}" ]; then
     java -jar NTPClient/build/libs/NTPClient-1.0-SNAPSHOT.jar "time.google.com" "123" "3" &
 fi
 
+# ntp_pass queries the NTP server learnt from DHCP. ntp_fail sends to time.google.com
+if [ -n "${options[ntp_pass]}" ] || [ -n "${options[ntp_fail]}" ]; then
```
grafnu, line 145: NB, can be written: 

[ -n "${options[ntp_pass]}" -o -n "${options[ntp_fail]}" ];

The current thing you have is ok, but is mixing boolean logic layers, essentially extra redundant parens.

docker/include/bin/start_faux
```diff
@@ -141,6 +141,26 @@ if [ -n "${options[ntp_client]}" ]; then
...
+        if [ -n "${options[ntp_pass]}" ]; then
+            dhcp_ntp=$(fgrep NTPSERVERS= /run/ntpdate.dhcp)
+            if [ -n "${dhcp_ntp}" ]; then
+                ntp_server=`echo $dhcp_ntp | cut -d "'" -f 2`
+                echo NTP server from DHCP $ntp_server
+            fi
+        elif [ -n "${options[ntp_fail]}" ]; then
+            ntp_server=time.google.com
+        fi
+        if [ -n "${ntp_server}" ]; then
```
grafnu, line 156: if statement is redundant, there is no case when it can get here without this being defined

testing/test_aux.out
```diff
@@ -5,14 +5,17 @@ check_style exit code 0
...
 RESULT skip base.startup.dns No dns traffic detected
 RESULT pass base.startup.dhcp
 RESULT skip base.switch.ping No local IP has been set, check system config
 RESULT pass base.target.ping target reached
+RESULT fail base.startup.ntp Invalid NTP server address
 RESULT fail base.startup.dns Invalid DNS server address
 RESULT pass base.startup.dhcp
 RESULT skip base.switch.ping No local IP has been set, check system config
 RESULT pass base.target.ping target reached
+RESULT fail base.startup.ntp Invalid NTP server address
```
grafnu, line 18: I'd either wait on this PR until one of these can be made skip, or file a bug for us to go back and clean it up.

docker/include/bin/start_faux
pbatta: updated, thanks.

docker/include/bin/start_faux
pbatta: updated, thanks.

testing/test_aux.out
pbatta: fixed it so we handle the skip case.

PR #496, opened 2020-06-22
USI

henry54809:


usi/src/main/java/switchtest/USIImpl.java
```diff
@@ -0,0 +1,92 @@
+package switchtest;
+
+import grpc.*;
```
grafnu, line 3: I don't think we ever converged on the "why grpc" question in the design proposal -- it seems to have been deleted without resolution. Also the design proposal was never finalized (still indicated as draft). I want to understand "why grpc" here before proceeding... it's a significant bit of complexity for something that could really be simple. (If we already had a conversation about it please remind me.)  I'm not saying it's *wrong*, I just want to make sure we're properly reviewing/discussing things before executing on them (it's a procedural thing).

usi/Dockerfile.usi
```diff
@@ -0,0 +1,12 @@
+FROM daqf/aardvark:latest
+
+# Do this alone first so it can be re-used by other build files.
+RUN $AG update && $AG install openjdk-9-jre
+
+RUN $AG update && $AG install openjdk-9-jdk git
+
+COPY usi/ usi/
+
+RUN cd usi && mvn clean compile assembly:single
```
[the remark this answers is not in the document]
henry54809: I prefer gradle as well actually.

usi/src/main/java/switchtest/USIImpl.java
henry54809: I remember seeing a comment related to grpc in my email but I couldn't find the comment when I went to the design doc. I don't know any good alternatives for this purpose?

grafnu:
It's in the document comment history.  Also should go through the process
of finalizing the doc (from draft --> (review) --> approved). We can close
it out there.  I'm mostly looking for a high-level description (in the
doc), about what the RPC is -- IIRC, it never really says what the contents
of the RPC are.

usi/src/main/java/daq/usi/SwitchController.java
```diff
@@ -0,0 +1,201 @@
...
+
+  public SwitchController(String remoteIpAddress, int telnetPort, String username,
+                          String password) {
+    this(remoteIpAddress, telnetPort, username, password, false);
+  }
+
+  /**
+   * Abstract Switch controller. Override this class for switch specific implementation
+   * @param remoteIpAddress switch ip address
+   * @param telnetPort  switch telnet port
```
grafnu, line 54: extra space

henry54809:
PTAL

Issue #336, opened 2020-03-31
Validate connection.min_send


spectre-storm:

noursaidi:
@pisuke @em-redstone , only comment on this test is that the report appendix states that "Device sends data at a frequency of less than 5 minutes", I think this should be more general as the test depends on the monitor_scan which is configurable within DAQ, and may not necessarily be 5 minutes.

pisuke:
@noursaidi @spectre-storm @em-redstone the best way would be to add a parameter in the module_config.json to be used by the test for the configuration of the monitoring time.
I think we need to discuss this with Trevor.

pisuke:
Addressed in #498

PR #498, opened 2020-06-23
Update Minimum Send Test

noursaidi:
Issue #336 

Update the minimum send test within the network test module to take into account the length of the monitor scan to check if a device sends any data packet at a frequency of less than five minutes. The changes are:
- Time between packets is measured - the test will pass if the time between any two is packets less than five minutes (deals with case where a monitor scan is long)
-  If no data packets are sent and the monitor scan period is short, the test will skip instead of failing


grafnu:
Code looks fine, but there's still some lint errors...

noursaidi:
@grafnu fixed the remaining lint errors. The print statement for the exception (in the event that 
 system.conf could not be found - which shouldn't happen) was changed to be written to the report.

pisuke:
@noursaidi thanks for this, I've merged. Please add both pass and fail instances in test_aux for this test in a separate PR.

PR #505, opened 2020-06-25
Issue337 multicastnetworktest

em-redstone:
Hi @pisuke @noursaidi @grafnu , please take a look at these recent changes,

Changelog:
- communication.type.broadcast now counts the number of multicast packets. It does this by checking if the addresses are within the range 224.0.0.0 to 239.255.255.255 (check high order bit on first address byte).
- Removed the debug_generate_capture.py as it just runs a simple tcpdump command to create a sample .pcap file. There are no code references to the file.
- Added some extra documentation.
- Fixed some typos. Did some slight code cleanup.
- Added network tests back into all.conf to run during test_aux.sh

Many thanks!


testing/test_aux.out
```diff
@@ -47,16 +47,31 @@ RESULT skip security.passwords.http Port 80 is not open on target device.
...
 RESULT skip security.passwords.telnet Port 23 is not open on target device.
 RESULT skip security.passwords.ssh Port 22 is not open on target device.
-RESULT fail security.passwords.http Default passwords have not been changed.
-RESULT fail security.passwords.https Default passwords have not been changed.
-RESULT fail security.passwords.telnet Default passwords have not been changed.
-RESULT fail security.passwords.ssh Default passwords have not been changed.
+RESULT skip security.passwords.http Port 80 is not open on target device.
+RESULT skip security.passwords.https Port 443 is not open on target device.
+RESULT skip security.passwords.telnet Port 23 is not open on target device.
+RESULT skip security.passwords.ssh Port 22 is not open on target device.
```
grafnu, line 53: Why this change?  What happened to testing the fail condition?

testing/test_aux.out
```diff
@@ -47,16 +47,31 @@ RESULT skip security.passwords.http Port 80 is not open on target device.
...
 RESULT pass security.passwords.http Default passwords have been changed.
 RESULT pass security.passwords.https Default passwords have been changed.
 RESULT pass security.passwords.telnet Default passwords have been changed.
 RESULT pass security.passwords.ssh Default passwords have been changed.
 RESULT skip security.firmware Could not retrieve a firmware version with nmap. Check bacnet port.
 RESULT pass security.firmware version found: ?\xFF\xFF\x19,>u\x08\x00no
+RESULT fail connection.dhcp_long 
+RESULT fail connection.min_send 
+RESULT info communication.type.broadcast 
+RESULT fail protocol.app_min_send
```
grafnu, line 63: why are all these fail?  Where's the tests for pass or skip?

em-redstone: Hi Trevor,

I've only tried to made code changes with regards to the communication.type.broadcast test - since there's been more frequent changes to the network module, I've added the network module back into test_aux.sh - but it's been removed for a long while so I'm not sure why the others are all failing or whether they were always like this, but I'll look into this.

testing/test_aux.out
em-redstone: I'll address this issue with the next commits - sorry, I completely missed this!

testing/test_aux.out
noursaidi: These tests rely on reading monitor_scan.pcap which is currently turned off within by test_aux.sh which might explain the fail. The setting might need to be changed

@em-redstone to note that connection.dhcp_long will need to be removed from the module (will form part of a different test module). The same for app_min_send as it's not part of the test plan

em-redstone: Thanks @noursaidi, I can remove these two tests in the next commit if that's what you meant, or did you mean to just make note of it in the readme.md?

noursaidi: @em-redstone I was noting it in case you were going to try and get the aux tests to pass/fail for them. We can do a seperate issue/PR to remove them

grafnu:
What we should do is separate out the problems into two categories...

   1. Something new that was inadvertently introduced, and shouldn't happen
   (making the system worse). I think this would manifest itself as changes in
   the test result from pass <-> fail etc...
   2. Something that was there before, but is now just being exposed. This
   would manifest itself as a new test result that, although wrong, wasn't
   there before. I don't want to shoot the messenger here.

For this PR, you should fix things that are in category #1. We don't want
things that were working before to break. But, you should NOT fix things in
#2 -- it's ok to expose them (that's good), but I don't want to bog this PR
down with fixing everything. We should then address that in another PR.

What we need to do then, is accurately triage the cases into #1 or #2.  It
sounds like one set is category #1, while the other set is #2.

em-redstone:
> What we should do is separate out the problems into two categories... 1. Something new that was inadvertently introduced, and shouldn't happen (making the system worse). I think this would manifest itself as changes in the test result from pass <-> fail etc... 2. Something that was there before, but is now just being exposed. This would manifest itself as a new test result that, although wrong, wasn't there before. I don't want to shoot the messenger here. For this PR, you should fix things that are in category #1. We don't want things that were working before to break. But, you should NOT fix things in #2 -- it's ok to expose them (that's good), but I don't want to bog this PR down with fixing everything. We should then address that in another PR. What we need to do then, is accurately triage the cases into #1 or #2. It sounds like one set is category #1, while the other set is #2.

Hi @grafnu, @noursaidi,

I've just compared the network module results in test_aux.out in a run of the master vs. multicast branch and they have the same results so I assume this was an issue that was present for some time only to be found out now #2.

I can address the issues related to the multicast test in this PR such as the lint issues and the result changes in test_aux.out #1, but for the failing network tests and removal of the two tests we could probably make another PR for those.

Regards

em-redstone:
Hi all,

I've fixed the lint errors, and the test_aux.out results are back to how they should be for security.password. I think it was due to the docker containers in my machine - I get these from time to time.

With the failing network module results, should I just remove them for now from test_aux for this PR?

Thanks.

testing/test_aux.out
grafnu: I think we should keep them as is (exposed) and then fix them in another PR.  Before they were just hidden which is the worst situation. Now we can at least see what is wrong!  I'll send a note to the team to figure out how to address this.

pisuke:
Closing this as it is addressed in #565

PR #520, opened 2020-06-30
dhcp class + using usi in daq

henry54809:


daq/gateway.py
```diff
@@ -216,19 +187,16 @@ def get_targets(self):
...
-        if self.dhcp_monitor:
-            self.dhcp_monitor.cleanup()
-        self._scan_finalize()
-        if self.host:
+        for _, dhcp_server in self._dhcp_servers.items():
             try:
-                self.host.terminate()
-                self.runner.remove_host(self.host)
-                self.host = None
+                dhcp_server.terminate()
```
grafnu, line 195: looks like you lost the runner.remove_host

daq/host.py
```diff
@@ -602,6 +641,7 @@ def _get_switch_config(self):
         return {
             'ip': self.switch_setup.get('ip_addr'),
             'model': self.switch_setup.get('model'),
+            'telnet_port': int(self.switch_setup.get('telnet_port', self._DEFAULT_TELNET_PORT)),
```
grafnu, line 644: 'telnet_port' isn't right -- what is it used for?  usi_port?

daq/host.py
```diff
@@ -5,9 +5,12 @@
 import shutil
 import time
 from datetime import timedelta, datetime
+import grpc
 
 from clib import tcpdump_helper
 from report import ResultType, ReportGenerator
+from proto import usi_pb2 as usi
+from proto import usi_pb2_grpc as usi_service
```
grafnu, line 13: it's feeling like much of this could/should be extracted to a separate usi module?

daq/host.py
```diff
@@ -59,6 +64,7 @@ class ConnectedHost:
     """Class managing a device-under-test"""
 
     _STARTUP_MIN_TIME_SEC = 5
+    _DEFAULT_TELNET_PORT = 23
```
grafnu, line 67: needs more qualifiers... maybe add switch in there?

proto/system_config.proto
```diff
@@ -114,6 +114,9 @@ message DaqConfig {
   
   // Set time between port disconnect and host tests shutdown
   int32 port_flap_timeout_sec = 48;
+
+  // USI url
+  string usi_url = 49;
```
grafnu, line 119: I would encapsulate this in a message -- it has a feeling that eventually it might have more parameters.

proto/system_config.proto
```diff
@@ -139,6 +142,9 @@ message SwitchSetup {
 
   // IP address template and subnet for module ip addresses
   string mods_addr = 16;
+ 
+  // Switch connect Telnet port
```
grafnu, line 146: I think this needs a slightly better name "telnet_port" is very generic... maybe cmd_port?  Talk about what it's use is, not the underlying protocol.

bin/usi
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash -e
 
 echo Starting USI 
-docker run -d --network=host --name daq-usi daqf/usi
+docker run -d --network=host --name daq-usi daqf/usi || true
```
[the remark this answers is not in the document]
henry54809: topo integration test doesn't need it so the image wasn't build for that test. I could either build this image for topo for ignore the failure. 

> Something's not right with the refactoring -- it doesn't match what was in the design proposal and I'm really not convinced there should be multiple dhcp containers.
The refactoring wasn't covered in the design proposal so which part were you referring to?

daq/dhcp_server.py
```diff
@@ -0,0 +1,130 @@
...
+            raise
+
+    def _initialize(self):
+        LOGGER.info('Initializing DHCP server on port %d', self._host_port)
+        cls = docker_host.make_docker_host(
+            'daqf/networking', prefix='daq', network='bridge')
+        # Work around an instability in the faucet/clib/docker library, b/152520627.
+        if getattr(cls, 'pullImage'):
+            setattr(cls, 'pullImage', lambda x: True)
+        host_name = 'dhcp%02d' % self._host_port
```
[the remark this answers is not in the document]
henry54809: The first DHCP container is the gateway. If there are additional DHCP servers, those will just be DHCP servers.

daq/gateway.py
henry54809: that's in dhcp_server's terminate

grafnu:
In the proposal diagrams all the DHCP servers are inside the Gateway
container, not separate containers.

I would build the container for the topo tests. I'd rather keep the hard
failure in case of an error.

daq/host.py
henry54809: the telnet port for usi to talk to the switch. usi port is defined in the usi_url config

grafnu: as I mentioned elsewhere "telnet_port" is way to generic and not all that interesting (telnet is a protocol) -- the name should indicate more the use not the protocol.

henry54809:
You mentioned that we should limit to 1 DHCP server per container so I thought the diagram was clear in that the DHCP servers are part of a gateway but are different containers; plus I don't think it's possible to run multiple dnsmasq on the same container since they will attempt to listen on the same port.  Original intention was for gateway to be its own container with multiple DHCP servers attached but I realized the gateway would be doing nothing at the point so I just made the first DHCP server the gateway container.

daq/host.py
henry54809: I understand that the _DEFAULT_TELNET_PORT should be more descriptive but the telnet_port is already under the switch_setup config so isn't saying switch_telnet_port redundant?  same with username, password, model etc.

grafnu:
Yes -- that is true. I think my reaction is something more about describing
the use, not the protocol... e.g. cmd_port, management_port, etc...

grafnu:
I don't think that's a safe assumption to make about the diagram. The proposal/diagram should be a very clear description of the relevant aspect. The gateway is a container, and the diagram shows DHCP servers in that container as the simplest explanation. This is the kind of complexity I was talking about when I said that I don't think we should be considering the multiple-DHCP servers if it complicates the architecture at all. You're complicating the (conceptual) architecture for something that's not needed.

The "gateway" is the higher-order concept than the DHCP server, since it's a container that does a bunch of different things _one_ of which is being a DHCP server, so it does not make sense to invert the relationship. If we need multiple servers in the future then we can cross that bridge when we come to it but for now, the basic structure of "the gateway" should remain constant.  It's not a big structural difference over what you have now, it's just organized slightly differently.

I would recommend you convert your new DHCPServer class to a GatewayHost class and then we can deal with the ramifications of multiple of them later. It's essentially keeping the refactoring smaller, so not changing the significant semantics (e.g. the gateway construct remains unchanged), but it's setting the groundwork for a different change in the future (e.g. the re-introduction of a DHCPServer or something).

daq/host.py
henry54809: I don't think it's really necessary at the moment. It brings more complexity but the logic to build these messages types still won't be contained in that usi wrapper class.

daq/host.py
```diff
@@ -285,6 +309,21 @@ def notify_activate(self):
             self._record_result('startup', state=MODE.HOLD)
         return self.state == _STATE.WAITING
 
+    def connect_port(self, connect):
```
henry54809, line 312: @grafnu here.

henry54809:
Updated to use USI OVS switch. PTAL.

bin/usi
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash -e
```
grafnu, line 1: I know it's weird and a bit off, but for now I think this should go in cmd/ and not bin/ -- simply because it's most similar to things like cmd/faux which do the same thing (start container).  At some point we have to clean everything up so it makes sense, but for now I'm going with consistency...

cmd/exrun
```diff
@@ -81,7 +82,8 @@ sudo rm -f $cleanup_file
 function autostart {
     tmp=`mktemp`
     echo DAQ autostart $@
-    eval $@ | tee $tmp
+    eval $@ > $tmp #Don't use "eval $@ | tee $tmp" here. Doesn't work for bin/usi
```
grafnu, line 85: space after #, and period at end of sentence.

usi/src/main/java/daq/usi/ovs/OpenVSwitch.java
```diff
@@ -31,7 +33,7 @@ protected String getInterfaceByPort(int devicePort) throws FileNotFoundException
       Matcher m = pattern.matcher(line);
       return m.find();
     }).findFirst().get();
-
+    bufferedReader.close();
```
grafnu, line 36: generall the java try-with-resources pattern is better than explicitly closing the object. https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

daq/host.py
```diff
@@ -271,6 +277,22 @@ def _state_transition(self, target, expected=None):
         LOGGER.debug('Target port %d state: %s -> %s', self.target_port, self.state, target)
         self.state = target
 
+    def _build_switch_info(self) -> usi.SwitchInfo:
+        switch_config = self._get_switch_config()
+        switch_model = switch_config["model"]
+        if switch_model in usi.SwitchModel.keys():
+            switch_model = usi.SwitchModel.Value(switch_model)
+        else:
+            switch_model = usi.SwitchModel.OVS_SWITCH
```
grafnu, line 286: seems like this should be a hard error not default.  What's the logic behind defaulting to OVS_SWITCH with when switch model is, e.g.,  ALIED_TELESIS_X230?

daq/gateway.py
```diff
@@ -33,97 +30,89 @@ def __init__(self, runner, name, port_set, network):
...
-            setattr(cls, 'pullImage', lambda x: True)
-        host = self.runner.add_host(host_name, port=host_port, cls=cls, tmpdir=self.tmpdir)
-        host.activate()
-        self.host = host
-        self._change_lease_time(self.runner.config.get('initial_dhcp_lease_time'))
-        LOGGER.info("Added networking host %s on port %d at %s", host_name, host_port, host.IP())
+        LOGGER.info('Initializing gateway %s as %s/%d',
+                    self.name, host_name, host_port)
+        self.tmpdir = self._setup_tmpdir('inst', host_name)
+        self.host = self._start_dhcp_server(host_port).host
```
grafnu, line 58: As per previous discussion, the top-level primary construct here should be the gateway (or "networking" container if you prefer). For the simple case, DHCP should be a sub-component of that.

henry54809:
Decided to try out a gateway container as the top level construct with separate dhcp server container(s). All changes related to that are contained in the last commit. I think the changes maintain the current structure but offer more flexibility without much complexity. There can be more optimization in terms of the starting of the DHCP server containers but I want to see what you think about this first. 
PTAL

grafnu:
I think the underlying problem is that you're conflating two different
concepts and so when you're trying to separate out multiple DHCP servers
it's breaking the abstraction in other layers of the (conceptual) stack.
Part of the problem I'll admit is that the terms "gateway" and "networking"
(container) are both ambiguous and unnecessarily different (e.g. it should
have been the networking class that runs the networking container...).  The
container really represents a host in the networking sense, which can
manage a bunch of different networking services (NTP, DNS, DHCP, NAT,
etc...).  So, for the simple case there should be one instance of this that
encapsulates all of the necessary networking services (including DHCP).
This is more or less a manifestation of something like my home network box
that does this stuff.  Key concept here is that DHCP is a service that runs
inside of this container.

Now comes the question of handling multiple DHCP servers, if/as/when
necessary. To do that, it might require spinning up an additional
(networking) container that *only* does DHCP and nothing else. So, the
property of the simple case (only one) doesn't complicate the system, but
it still allows for multiple server hosts if/as necessary. Conceptually,
this is the same networking services image, it just happens to be invoked
in a way that only does DHCP and not the other stuff. Generalizing, there
might be some other redundant service as well and this would provide a
container that could also provide that (e.g. a host can be configured to
have multiple DNS servers).

I think the problem can be distilled down to the name of the
"_start_dhcp_server" method -- which is really starting a dhcp container
(not just server). It should be something like
"_start_networking_host(secondary=True)" or something like that. And then
the first instance would be just "_start_networking_host()".  So, from a
container perspective there's only one kind (networking services), and the
first instance runs all sorts of stuff.  And non-primary instances of it
are more or less the same just running a diminished set of services.

I would recommend starting off by trying to diagram this cleanly -- just
one set of symbols for containers, and then a representation of the
services that run inside of those containers... would be a lot easier to
diagram and ratify before refactoring the code.  And then there's two
versions -- one for the simple case that just supports one DHCP server, and
then another that supports two. The problem we ran into here is your
original diagram was ambiguous (given that we interpreted it differently)
and that disconnect never got addressed before moving on to code.

daq/dhcp_server.py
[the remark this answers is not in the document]
grafnu: I think this statement also is a manifestation of the confusion... it should read:

The first networking container is the gateway (including DHCP). If there are additional DHCP servers, those will be in limited-function networking containers.

You had up-leveled the concept of "DHCP server" to that of "DHCP container" which is where the problem started (IMHO).

henry54809:
I think it's best to keep the original gateway class for now. I've reverted changes related to gateway and dhcp server. Spawning more networking containers(separate from gateway or not) does bring more complexity with the current ntp setup which will take more time that we can afford right now to resolve. We can revisit this when we need to convert gateway for the orchestrated testing / support multiple dhcp servers. PTAL

usi/src/main/java/daq/usi/ovs/OpenVSwitch.java
```diff
@@ -11,61 +11,64 @@
...
-    Matcher m = pattern.matcher(interfaceLine);
-    m.matches();
-    return m.group(3);
+    try (BufferedReader bufferedReader = new BufferedReader(reader)) {
+      Pattern pattern = Pattern.compile("(^\\s*" + devicePort + ")(\\((.+)\\))(:.*)", 'g');
+      String interfaceLine = bufferedReader.lines().filter(line -> {
+        Matcher m = pattern.matcher(line);
+        return m.find();
+      }).findFirst().get();
+      bufferedReader.close();
```
grafnu, line 36: close is handled implicitly by the try clause, so you shouldn't do it explicitly here

PR #522, opened 2020-07-03
Add DAQ version to origin summary

grafnu:


firebase/functions/index.js
```diff
@@ -193,17 +193,22 @@ function handleTestResult(origin, siteName, message) {
...
+    }),
     heartbeatDoc.get().then((result) => {
       const current = result.data();
-      if (!current || !current.message || current.message.timestamp < message.timestamp)
+      const defined = current && current.message && current.message.timestamp;
+      if (!defined || current.message.timestamp < message.timestamp) {
         return heartbeatDoc.set({
           'updated': timestamp,
-          message
+          'message': message
```
henry54809, line 209: {message} works the same as {'message': message} in js.

grafnu:
Restored. I think I half-figured that out when I saw it but it rubs me the
wrong way (to me it's conflating the variable name with the object key,
which aren't the same thing!) -- but -- put it back so as not to change
things just for the sake of changing things.

Issue #510, opened 2020-06-26
Develop test to identify SSH version for a device


noursaidi:

noursaidi:
PR #523 implementing this test merged

PR #523, opened 2020-07-06
add security.ssh.version test 

noursaidi:
Issue #510 - adds a new test which checks if an SSH server is running, and if so that it is SSHv2


noursaidi:
@grafnu @pisuke could you please review this test and let me know of any comments you may have? The Travis test passed on my run (https://travis-ci.com/github/noursaidi/daq/builds/174523051), and seems to be a dhcp/ntp issue. Thank you

docker/include/bin/start_faux
```diff
@@ -126,6 +126,12 @@ if [ -n "${options[telnet]}" ]; then
     (while true; do echo Telnet `hostname`; nc -nvlt -p 23 -e `which hostname`; done) &
 fi
 
+# SSH server is installed but not running
```
grafnu, line 129: What do you mean by "not running"?  This code explicitly starts the server... ?

subset/security/build.conf
```diff
@@ -1,3 +1,4 @@
 build subset/security
 add tls
 add password
+add ssh
```
grafnu, line 4: missing newline

testing/test_aux.out
```diff
@@ -57,6 +57,9 @@ RESULT pass security.passwords.telnet Default passwords have been changed.
 RESULT pass security.passwords.ssh Default passwords have been changed.
 RESULT skip security.firmware Could not retrieve a firmware version with nmap. Check bacnet port.
 RESULT pass security.firmware version found: ?\xFF\xFF\x19,>u\x08\x00no
+RESULT pass security.ssh.version Device only supports SSHv2
```
grafnu, line 60: There should be { pass, skip, fail } coverage.
You should consider putting these in test_modules.sh -- it's a bit quicker to setup and develop than test_aux.sh

noursaidi: Thanks @grafnu, I've added coverage for pass, skip and fail. I've also removed the tests from test_aux and added them in to test_modules

docker/include/bin/start_faux
noursaidi: That was a note to myself which should have been removed - you're right that's exactly what it does. Removed now.

noursaidi:
@grafnu thank you for taking a look and providing feedback, I've now addressed these. Could you please let me know if you have any comments on the changes

docker/modules/Dockerfile.faux1
```diff
@@ -18,13 +18,50 @@ RUN bin/retry_cmd git clone https://github.com/grafnu/bacnet4j.git --single-bran
 COPY pubber/ pubber/
 RUN pubber/bin/build
 
+# Dockerfile to build older version of SSH and SSL
+FROM daqf/aardvark:latest as ssh_build
+
+RUN $AG update && $AG install wget make build-essential gcc libz-dev ca-certificates
+
+# Build SSH, OpenSSL from source and configure
+RUN wget https://www.openssl.org/source/openssl-1.0.2g.tar.gz && \
```
grafnu, line 27: Can you put this and the next RUN command as scripts in the appropriate directory?  This is an awful lot of stuff to have inlined in the Dockerfile.

noursaidi: @grafnu I've split them out into separate files as suggested

PR #531, opened 2020-07-13
Support for alternate sec switch (not managed by DAQ)

grafnu:


config/system/alt.yaml
```diff
@@ -0,0 +1,20 @@
+# Example configuration file for using an OVS switch not managed by DAQ.
+
+# Load defaults.
+include: config/system/default.yaml
+
+# Description description for dashboard.
```
pbatta, line 6: cosmetic: 's/Description description/Description/'

grafnu: Updated, PTAL.

PR #532, opened 2020-07-13
Additional DHCP test part 1

henry54809:


daq/host.py
```diff
@@ -50,11 +51,22 @@ class MODE:
     MERR = 'merr'
 
 
+# Mapping DHCP test name to host function call
+DHCP_TESTS_MAPPING = {
+    'port_toggle': '_dhcp_port_toggle_test'
```
grafnu, line 56: I think this would be better served as a lambda rather than string-converted-to-attribute, e.g., something like https://github.com/grafnu/forch/blob/master/forch/forchestrator.py#L299

It can easily be dynamic (rather than a STATIC_MAPPING) -- performance is not an issue here...

daq/host.py
```diff
@@ -556,13 +582,28 @@ def _base_tests(self):
...
+        self._ip_listener = ip_listener
+
     def _run_next_test(self):
         try:
             if self.remaining_tests:
                 LOGGER.debug('Target port %d executing tests %s',
                              self.target_port, self.remaining_tests)
                 self.timeout_handler = self._main_module_timeout_handler
                 self._docker_test(self.remaining_tests.pop(0))
+            elif self.dhcp_tests:
```
grafnu, line 605: I'm not sure exposing dhcp_tests is top-level construct is quite the right way to go here. I would have thought a "dhcp_test" would be a sub-class of a "inline_test" which is a subclass of a "module" (not a literal sub-class, but as a general concept). Going forward, there might be other kinds of inline tests that aren't DHCP, and there's going to be native tests that are modules but not docker tests.  Right now, there's a conflating of "docker_test" and "module" -- but that's just a collapsed abstraction and I'd move in the direction of pulling that apart.   I think the problem in the code here is actually the line above this (line 604), where "remaining_tests" maps 1:1 to "dhcp_tests", which is where I think it'd be better served to have the right abstraction in place.

If we don't know about the implementation, is there anything fundamentally different about the dhcp_tests from the other tests?

resources/setups/baseline/module_config.json
```diff
@@ -36,5 +36,11 @@
     "ntp": {
       "enabled": true
     }
+  },
+  "dhcp_tests": {
```
grafnu, line 40: as per other comments, I'm not sure this should be a top-level construct at the same level as "modules" -- I would have expected "modules" to contain a "dhcp" module...

daq/host.py
```diff
@@ -194,6 +212,10 @@ def _get_static_ip(self):
     def _get_dhcp_mode(self):
         return self._loaded_config['modules'].get('ipaddr', {}).get('dhcp_mode', 'normal')
 
+    def _get_dhcp_tests(self):
+        tests = self._loaded_config.get('dhcp_tests', {}).keys()
```
grafnu, line 216: A related fundamental question I have is if "ipaddr" and "dhcp_tests" are conceptually related. They seem to fall into the same conceptual bucket (things that relate to the ipaddress of the device) so wondering if there's a cleaner consolidation that streamlines things?

daq/host.py
```diff
@@ -556,13 +582,28 @@ def _base_tests(self):
...
+    def _dhcp_port_toggle_test(self, logging_handler):
+        def ip_listener(target_ip):
+            LOGGER.info("%s test Received ip: %s" % (self.test_name, target_ip))
+            if logging_handler:
+                LOGGER.removeHandler(logging_handler)
+            self._ip_listener = None
+            self._end_test()
+
+        self.connect_port(False)
+        time.sleep(5)
```
grafnu, line 594: 5 should be configurable (w/ default) in some form.

daq/host.py
henry54809: i considered that but I wanted to reuse the keys of DHCP_TESTS_MAPPING to generate dhcp_tests() function so you don't have specify twice for the available DHCP tests.

grafnu:
They are not mutually exclusive -- you can create a map of lambda
functions rather than a map of strings -- the keys are the same in both
cases.  (I wasn't talking about the "register" pattern, just the use of
lambdas rather than string values)

henry54809:
The lambas need to reference each host instance with the self var; how do you access self outside of the host class? If you move the map into the host class, than the dhcp_tests fn that's outside of the host class can't reference the map.

grafnu:
Then move the dhcp_tests inside of the host class?  What's the core
fundamental reason this has to be static?  Also, conceptually this is just
one module -- so this is a lot of excess scaffolding just to support one
thing since there's no generalization. Seems like this would all be a lot
easier with a simple if statement in the right place.

henry54809:
PTAL

PR #536, opened 2020-07-15
Refactor ipaddress module

grafnu:
Not done yet (still need to get all the tests to pass... aka make it actually work), but this is the base of the refactoring.

### DIFF
--- a/daq/host.py
+++ b/daq/host.py
@@ -300,11 +300,11 @@ class ConnectedHost:
 
     def _build_switch_info(self) -> usi.SwitchInfo:
         switch_config = self._get_switch_config()
-        switch_model = switch_config['model']
-        if switch_model == 'FAUX_SWITCH':
+        model_str = switch_config['model']
+        if model_str == 'FAUX_SWITCH':
             return None
-        else if switch_model:
-            switch_model = usi.SwitchModel.Value(switch_config["model"])
+        if model_str:
+            switch_model = usi.SwitchModel.Value(model_str)
         else:
             switch_model = usi.SwitchModel.OVS_SWITCH
         params = {
--- a/daq/host.py
+++ b/daq/host.py
@@ -300,7 +300,10 @@ class ConnectedHost:
 
     def _build_switch_info(self) -> usi.SwitchInfo:
         switch_config = self._get_switch_config()
-        if switch_config["model"]:
+        switch_model = switch_config['model']
+        if switch_model == 'FAUX_SWITCH':
+            return None
+        else if switch_model:
             switch_model = usi.SwitchModel.Value(switch_config["model"])
         else:
             switch_model = usi.SwitchModel.OVS_SWITCH
@@ -330,6 +333,9 @@ class ConnectedHost:
     def connect_port(self, connect):
         """Connects/Disconnects port for this host"""
         switch_info = self._build_switch_info()
+        if not switch_info:
+            self.logger.info('No switch model found, skipping port connect')
+            return False
         try:
             with grpc.insecure_channel(self._usi_url) as channel:
                 stub = usi_service.USIServiceStub(channel)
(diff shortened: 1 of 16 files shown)


henry54809:
I thought the Abstraction is for any test under a module rather than specifically for ip addr test.

grafnu:
I'm not sure what you mean. There's two things going on -- the abstraction
itself which is only loosely applied (since e.g. there is no actual base
class) and then the singular instantiation of it for ipaddr tests.

henry54809:
So a test class just like the existing docker test class rather than the current ipaddr test class just for IP addr test. Do you plan on having a different class for each existing test module, ping, switch nmap, etc?  
> I'm not sure what you mean. There's two things going on -- the abstraction itself which is only loosely applied (since e.g. there is no actual base class) and then the singular instantiation of it for ipaddr tests.
> […](#)

grafnu:
I would imaging the (virtual) module class hierarchy something like:

   - (TestModule)
      - Docker
      - (Native)
      - (Inline)
         - ipaddr

The things in () don't currently exist as actual code -- either because of
a collapsed abstraction (joys of ducktyping) or because they haven't been
written yet (Native). So, something like ping would not have a specific
class as long as it's instantiated as a docker container (how it currently
is) or a native module.  Right now, I'm not aware of anything else that
would be an "inline" module.  A native module would be very similar to a
docker module, in that the actual DAQ code for it would be only a wrapper
around an externally executed script.

usi/src/main/proto/usi.proto
```diff
@@ -36,6 +36,7 @@ enum SwitchModel {
   ALLIED_TELESIS_X230 = 0;
   CISCO_9300 = 1;
   OVS_SWITCH = 2;
+  FAUX_SWITCH = 3;
```
henry54809, line 39: what's this for?

grafnu: There's a setup where there is no switch that can be controlled (it's OVS but not reachable)... and that's what this field indicated.  but, this is not the right place to handle that, so I put a check in at a different layer to detect that connection and not allow the connect RPC call at all.

daq/host.py
```diff
@@ -629,10 +615,15 @@ def _device_aux_path(self):
             os.makedirs(path)
         return path
 
-    def _docker_test(self, test_name):
-        self.test_host = docker_test.DockerTest(self.runner, self.target_port,
-                                                self.devdir, test_name)
-        self.logger.debug('test_host start %s/%s', test_name, self._host_name())
+    def _new_test(self, test_name):
+        clazz = ipaddr_test.IpAddrTest if test_name == 'ipaddr' else docker_test.DockerTest
```
henry54809, line 619: So IpAddrTest is a single port toggle test inside the ipaddr module whereas the DockerTest is a module, since Host class should be host of modules, something is not aligned.  I think there should also be the test abstraction and module abstraction(other than dockertest) in place first before the current port_toggle test class. 
Ip addr test and DockerTest are also both named test which is why we should also spend some time clearly distinguish test and module in code when we have time for more refactoring.

grafnu: IpAddrTest and DockerTest are both "modules" in the sense they are a high-level entity that's managed by ConnectedHost. Both IpAddrTest and DockerTest, in turn, encapsulate individual line-item tests.  There is an abstraction in place, but maybe you mean that it should be an _explicit_ abstraction (e.g. with a named base-class or similar?)

grafnu:
Re: names.  Yes, they are not right and should be fixed, but that doesn't
affect the underlying abstractions. There's a distinction between a python
module and a test-module (class that's instantiated and managed by
ConnectedHost)... I think you're conflating them when you say "IpAddrTest
is a single port toggle test inside the ipaddr module" -- There's the
python module ipaddr_test.py, containing the test-module "class
IpAddrTest", and then there's line-item tests inside of that (function
_dhcp_port_toggle_test).

PR #539, opened 2020-07-16
Additional DHCP test part 2. Multisubnet test

henry54809:


docker/include/networking_scripts/change_dhcp_range
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash -e
+#
+# Dynamically change DHCP lease range, requires killing and restarting
+# dnsmasq as per documentation (SIGHUP does not reload configuration file).
+LOCAL_IF=${LOCAL_IF:-$HOSTNAME-eth0}
+
+range_start=$1
+range_end=$2
+net_mask=$3
```
grafnu, line 9: is this really net_mask?  Isn't the value "16"?  That's not a netmask, that's a CIDR or some kind of prefix space.. "net_mask" typically refers to the full 255.255.0.0 form.

testing/test_aux.out
```diff
@@ -94,7 +94,8 @@ port-01 module_config modules
     "enabled": false
   },
   "ipaddr": {
-    "enabled": true,
+    "dhcp_ranges": [["192.168.0.1", "192.168.255.254", 16]],
+    "enabled": false,
```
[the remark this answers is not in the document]
henry54809: yes. these are enabled only in test_many. It takes way too long to run these for them to be enabled by default.

docker/include/networking_scripts/change_dhcp_range
henry54809: What should it be named? prefix?

grafnu:
"The number following the slash is the prefix length"
https://en.wikipedia.org/wiki/Classless_Inter-Domain_Routing#CIDR_blocks --
so, likely prefix_len

henry54809:
Renamed to prefix_len. PTAL

PR #546, opened 2020-07-20
Move ntp test to monitor pcap

pbatta:


docker/include/bin/start_faux
```diff
@@ -147,8 +148,8 @@ if [ -n "${options[ntpv4]}" ]; then
     done) &
 elif [ -n "${options[ntpv3]}" ]; then
     (while date; do
-    	echo Transmitting NTP query to time.google.com using NTPv3
-    	ntpdate -q -o 3 time.google.com
+    	echo Transmitting NTP query to 216.239.35.8 using NTPv3
```
grafnu, line 151: Nit: consider extracting this to a named constant (e.g. STATIC_NTP_SERVER)... just reads a little bit better :-)

pbatta: done, thanks.

PR #548, opened 2020-07-21
USI not to use background shell

grafnu:


usi/src/main/java/daq/usi/ovs/OpenVSwitch.java
```diff
@@ -19,7 +19,7 @@
 
 public class OpenVSwitch implements SwitchController {
 
-  private static final String OVS_OUTPUT_FILE = "ovs_output.txt";
+  private static final String OVS_OUTPUT_FILE = "sec.ofctl";
```
henry54809, line 22: Also can you please change the test filename? usi/src/test/resources/ovs_output.txt

grafnu: Do you know why the test didn't fail?  Something like that should have been caught, so seems like something's misconfigured.

henry54809: The unit tests aren't setup to run but they will be as part of the github actions migration.

grafnu: Ok, renamed test resource... PTAL.

PR #555, opened 2020-07-21
GRPC timeouts + usi first command wait fix.

henry54809:


config/system/default.yaml
```diff
@@ -44,3 +44,4 @@ topology_hook: bin/dump_network
 # usi url for DAQ to connect to
 usi_setup: 
   url: localhost:5000
+  grpc_timeout_sec: 10
```
grafnu, line 47: just rpc_timeout_sec -- there's no need to bleed the implementation of "this is grpc"...

henry54809: renamed. PTAL

PR #557, opened 2020-07-23
Automatic build script

grafnu:


bin/build_release
```diff
@@ -0,0 +1,80 @@
...
+    false
+fi
+
+git checkout master
+
+changed=`git diff --name-only release_stable docs/changelog.md`
+if [ -z "$changed" ]; then
+    git log release_stable..HEAD --pretty=oneline | sed -e 's/[a-z0-9]+/\*/g'
+    echo docs/changelog.md has not been updated since last release_stable
+    echo Use the log lines above for inspiration.
```
pbatta, line 26: nice touch!

bin/build_release
```diff
@@ -0,0 +1,80 @@
...
+ M etc/docker_images.ver
+EOF
+
+git status --porcelain > /tmp/git_status.txt
+
+if ! diff /tmp/git_expected.txt /tmp/git_status.txt; then
+    echo Expected build images not found. Something went wrong.
+    false
+fi
+
```
pbatta, line 58: nit: maybe delete /tmp/git_expected.txt and /tmp/git_status.txt as they are not needed beyond this point

bin/build_release
```diff
@@ -0,0 +1,80 @@
...
+git commit -a -m "$VERSION release"
+git tag -a $VERSION -m "$VERSION release"
+git push
+git push --tags
+
+faucetgit=`git config remote.faucet.url`
```
pbatta, line 64: This would be applicable if the users fork of the repo is set up as a remote called 'faucet'? If so, maybe add a comment to that affect?

grafnu:
Updated, PTAL

PR #558, opened 2020-07-24
Github actions

henry54809:


bin/setup_base
```diff
@@ -61,7 +61,7 @@ echo Adding username to docker group...
 sudo groupadd docker || true
 sudo usermod -aG docker $user
 
-DEF_IFACE=`sudo route -n | fgrep UG | awk '{print $8}'`
+DEF_IFACE=`sudo route -n | fgrep UG | head -1 | awk '{print $8}'`
```
grafnu, line 64: We need to make sure this is the right logic.  How is it happening that there's two gateways?  The behaviour in this situation is undefined (which one is right?) so it usually an indicator of a misconfigured system.

bin/test_daq
```diff
@@ -13,7 +13,7 @@ function delay_finish {
     sleep 10
 }
 
-if [ -n "$TRAVIS" ]; then
+if [ -n "$CI" ]; then
```
grafnu, line 16: Where is CI defined?  Is that defined as part of GitHub actions?  If possible, I'd prefer something longer than "CI" (short names are always problematic), but if it's dictated from GitHub then it'll have to do...

henry54809: Yes, it's one of the github actions default env vars. https://docs.github.com/en/actions/configuring-and-managing-workflows/using-environment-variables#default-environment-variables

bin/setup_base
henry54809: updated. PTAL

subset/cloud/test_udmi
```diff
@@ -23,8 +23,9 @@ device_id=`jq -r .device_id /config/device/module_config.json`
 
 # Do basic network connectivity check
 cat /etc/resolv.conf
-ping -c 2 172.217.164.106
-ping -c 2 pubsub.googleapis.com
+ping -c 2 172.217.164.106 || true
+ping -c 2 pubsub.googleapis.com || true
+(curl pubsub.googleapis.com 2>&1 > /dev/null  2>&1 && echo "Can reach pubsub.googleapis.com") || echo "Can't reach pubsub.googleapis.com" && false
```
grafnu, line 28: Doesn't this return false in all cases?  Either way, maybe better to add some ()'s or whever the moral equivalent is here just to be clear what the logic is.

subset/cloud/test_udmi
```diff
@@ -23,8 +23,9 @@ device_id=`jq -r .device_id /config/device/module_config.json`
 
 # Do basic network connectivity check
 cat /etc/resolv.conf
-ping -c 2 172.217.164.106
-ping -c 2 pubsub.googleapis.com
+ping -c 2 172.217.164.106 || true
```
grafnu, line 26: I assume this is because ping or something is blocked on GitHub actions?  If so, this is fine, but maybe add a comment explaining why it's like this.

henry54809:
Updated docs. PTAL

README.md
```diff
@@ -1,5 +1,4 @@
-![DAQ unit test](https://github.com/faucetsdn/daq/workflows/DAQ%20unit%20test/badge.svg?branch=master)
-![DAQ integration tests](https://github.com/faucetsdn/daq/workflows/DAQ%20integration%20tests/badge.svg?branch=master)
+![DAQ unit test](https://github.com/faucetsdn/daq/workflows/DAQ%20main%20tests/badge.svg?branch=master)
```
grafnu, line 1: isn't this all DAQ tests now, not just unit tests?

docs/add_test.md
```diff
@@ -124,9 +124,9 @@ Tests should:
 - Include the test name and a description of the test in the report output
 - Include an informative line in the summary table
 
-Integration tests don't need to be tedious and, if you're developing one test and seeing a consistent failure on Travis CI, isolate your problem and run _just that part_ of the integration test both locally and on Travis CI.
+Integration tests don't need to be tedious and, if you're developing one test and seeing a consistent failure on [Github actions](https://github.com/faucetsdn/daq/actions), isolate your problem and run _just that part_ of the integration test both locally and on [Github actions](https://github.com/faucetsdn/daq/actions).
```
grafnu, line 127: nit: in my opinion it's probably overkill to link all of them since they're all the same. Just like when a new term is introduced using _italics_, it's only done the first time.

subset/cloud/test_udmi
```diff
@@ -5,13 +5,20 @@ source reporting.sh
 REPORT=/tmp/report.txt
 LOG=/tmp/udmi.log
 
+function delay_finish {
+    # Travis doesn't always wait for buffer to flush on exit, so give some time.
```
grafnu, line 9: nit: no more travis. Not sure if this is even necessary anymore, but doesn't hurt.

testing/test_aux.sh
```diff
@@ -122,8 +122,8 @@ sort inst/result.log | tee -a $TEST_RESULTS
 
 # Show partial logs from each test
 head inst/gw*/nodes/gw*/activate.log
-head inst/run-port-*/nodes/*/activate.log
-head inst/run-port-*/nodes/*/tmp/report.txt
+cat inst/run-port-*/nodes/*/activate.log
```
grafnu, line 125: restore to head

PR #559, opened 2020-07-28
Remove possibly deprecated test modules

grafnu:


pisuke:
@grafnu @noursaidi I think we can close this PR without merging as it is addressed in #565

PR #565, opened 2020-07-29
Cleanup Network Tests and Network Test Module

noursaidi:
PR to clean up the network tests and network test module. Summary of changes:
- Remove depreciated code from current network module
- Merged pull request #505 (multicast network test)
- Re-added network module into Travis CI test.
  - the min_send test always passes and the info_broadcast always the same message for now. I'm not sure if it's possible to test fail/skip as any network traffic from the faux device triggers a pass.
- Remove NTP test module and `subset/ntp` combined into existing network module
- Remove mac_oui test module and `subset/connection` directory, combined into existing network module


docker/modules/Dockerfile.faux1
```diff
@@ -34,12 +34,12 @@ FROM daqf/aardvark:latest
...
 RUN $AG update && $AG install openjdk-8-jdk git
 RUN $AG update && $AG install isc-dhcp-client ethtool network-manager netcat curl\
-    python ifupdown openssl nano apache2-utils ntpdate 
+    python ifupdown openssl nano apache2-utils ntpdate vzctl
 
 # Additional OS dependencies
 RUN $AG update && $AG install -y telnetd && $AG install xinetd nginx
 
-COPY subset/ntp/NTPClient NTPClient
+COPY subset/network/NTPClient NTPClient
```
grafnu, line 42: this shouldn't still be in subset/ -- it should be moved to docker/include/network

there should be no dependency on the subset/ folder from faux

testing/test_aux.sh
```diff
@@ -63,11 +63,11 @@ site_path: inst/test_site
 schema_path: schemas/udmi
 interfaces:
   faux-1:
-    opts: brute broadcast_client ntpv4
+    opts: brute broadcast_client ntpv4 arp
```
grafnu, line 66: why is this enabled for everything?  Whatever it's for, that's not the right way... If it's truly always required then it should be the default behavior, otherwise all three of these should not have it (what's the point?)

docker/modules/Dockerfile.faux1
noursaidi: @grafnu I was planning to address the dependancy as a separate PR as it involved other test modules too. I'll update to pick up the network module changes here

testing/test_aux.sh
noursaidi: Ah, that would have been a much better, thanks. This is related to the min_send test which always passes, but might have a different summary depending if there are any ARP packets. As there seems to be some chance it might not, this periodically sends ARP packets so the result is always as expected.

docker/include/bin/start_faux
```diff
@@ -193,16 +193,6 @@ if [ -n "${options[discover]}" ]; then
     bin/bacnet_discover loop &
 fi
 
-if [ -n "${options[curl]}" ]; then
```
grafnu, line 196: curl should still be optional since it's not default behavior for every case

noursaidi:
Thanks @grafnu for reviewing again, I'd appreciate your input because I'm not sure what the best way to proceed is. 

The issue I'm trying to resolve is a an unpredictable result summary in the Travis report output from the general network tests (communication_type and min_send).  

From 7 test_aux runs I've done, 2 have failed . One was due to one faux device not sending ARP packets, the other not sending unicast packets.  This respective changed the report summary by omitting a few words. I may be wrong, but I don't think this can be predicted.

Therefore I think 'arpsend' and 'curl' (or alternative to send a data packet) is needed as default behavior for test_aux only, but not for the faux devices generally, to ensure predictable behavior.

Alternatively, as these tests rely on reading monitor.pcap, For 2/3 of the faux devices I could add pre-prepared packet capture files and configure the test module to read that instead of the actual monitor scan. That way can at least have some test of skip/fail conditions. The test would still run normally on one of the faux devices. Not ideal, might be better than three tests with the same result.

Do you think there is a better way?

pisuke:
@grafnu does @noursaidi 's suggestion sound good in the short term to get this PR merged?

noursaidi:
@pisuke @grafnu I was thinking that instead I could add a configuration to enable/disable the sub tests from module_config, and disable the problematic general network tests in the Travis CI tests. This means the coverage of testing in Travis CI remains identical after this PR as it did before.  

I do think it is important to have proper coverage of all test, and then this can be discussed and addressed properly in a separate issue/PR rather.

grafnu:
For the short term, I'd make the arp send a default for faux. It's not
"wrong" but having it always on in faux should make it the easiest to clean
up. At the same time we should file a bug/issue to get it (arp being the
default) cleaned up.

Module configs are bad because then it's just making everything more messy
to work around this one case.

Doing pre-prepared capture packets correctly is likely harder than you
think, so I wouldn't go down that route. (Basically you can't closely tie
it in with the test itself, which means you need some mechanism to mock
them out, which means more work than you expect.)

The likely solution is to increase the monitoring time, but I'm not sure.
That would just be my guess about a flaky connection.

I don't understand why making 'curl' default should be necessary. But for
now, can you just make arp the default and we'll go from there.  And I
don't mean "arp is enabled on everything" like you had before, but more
that 'arp' is the default and there would be an explicit option (xarp) to
NOT do it.

noursaidi:
Really appreciate your thoughts and feedback @grafnu, thanks. I've restored curl and made xarp an option as as suggested.

Here is a run of 12 test_aux runs, passes 11. https://travis-ci.com/github/noursaidi/daq/builds/178423661.

'curl' default was overkill, only faux-3 needed something to periodically send data. The other 2 had NTP which satisfies this.

docker/include/bin/start_faux
```diff
@@ -224,12 +224,16 @@ if [ -n "${options[pubber]}" ]; then
...
+# Periodically sends ARP packets
+if [ -z "${options[xarp]}" ]; then
+    echo Starting arp send loop.
+    (while true; do arpsend -D -e 10.20.254.254 $intf_name; sleep 2; done) &
+fi
 
+if [ -n "${options[curl]}" ]; then
+    echo Starting curl loop.
+    (while true; do curl -o - http://google.com; sleep 1; done) &
+fi
```
grafnu, line 236: nit: blank line after this block

grafnu:
Great -- 11/12 passing is fine. If there's still flakiness then we can deal
with it separately.  PR approved with just a minor formatting comment.

Issue #560, opened 2020-07-29
Faux dependencies in subset folder


noursaidi:
Currently faux devices have dependencies in subset. These will need to be moved elsewhere. There are two types of dependencies: 
- Files used used to build the container
- Files which run or are used within the container. 

These are the following
- `subset/security/sshfaux/` - Contains 2 scripts used to build docker container only
- `subset/ntp/NTPClient/*`- Java Source for NTP client running on faux container
- `subset/network/TransportClient/*` - Python file running on faux container
- `subset/security/nginx-site/*` 
- `subset/security/nginxpass.conf ` - Configuration 
- `subset/security/nginxfail.conf` - Configuration 
- `subset/bacnet`
- `subset/pentests` - Bruteforce test (is this being retained?)

@grafnu @pisuke do you have any thoughts or comments on where these could go?

I think  `/docker/includes/root` which would effectively become a mirror of the `root` directory in the docker container for all the files. But I'm not sure where scripts used in the build process could go.

grafnu:
@pbatta I think docker/includes/security would be good for the subset/security ones, and then similar for docker/includes/network -- and then I'd put subset/ntp/NTPClient into docker/includes/network as well

noursaidi:
@grafnu sorry, I noticed I missed out two from my original list. These were: `subset/bacnet` and `subset/pentests`. The latter is for the bruteforce test (which might be removed?)

Are you still happy with these all moving into docker/includes as per the folder structure mentioned?

grafnu:
If they are used by the faux containers then they should be in docker/includes -- if they are used by the runtime tests then they should be in subset/

noursaidi:
PR #567 and #565 (network faux dependancies only) merged

PR #567, opened 2020-07-30
Remove faux dependencies from subset directory

noursaidi:
Issue #560 - dependencies for faux containers moved from subset into docker/includes directory.


noursaidi:
The 'new' files are duplicates of existing files, I will address the stickler errors in both the original (in subset) and the duplicate (in docker/include) files.

PR #572, opened 2020-08-03
Test infrastructure cleanup

grafnu:


henry54809:
I think there will be more USI tests which is why I left it in a separate workflow to make them easily distinguishable. Also the badge on README needs to be changed if you delete the other workflow.

grafnu:
In the grand scheme of things USI is only a small part of DAQ so I don't
think warrants a separate top-level entry b/c it significantly complicates
the UI.  If things reach the point where it makes sense to break it out
later we can, but for the scope it is now it should be consolidated.

PR #576, opened 2020-08-05
Adding default dns for static ip faux devices

henry54809:


docker/include/bin/start_faux
```diff
@@ -87,7 +87,23 @@ if [ -n "${options[xdhcp]}" ]; then
...
+        route add -net 172.17.0.0/16 $intf_name  # default docker ip range.
+        dhcp_dns=8.8.8.8
+        echo nameserver ${dhcp_dns#DNS=} > /etc/resolv.conf
+        # There is no deterministic way to find out the default gateway ip where running more than 1 faux device.
+        (
+            while true; do
+                if [ `route -n | grep UG | wc -l` -eq '1' ]; then 
+                    timeout 1 ping -c 1 `route -n | grep UG ` && continue
+                fi 
+                for IP in {1..12}
```
grafnu, line 100: do on same line as for

docker/include/bin/start_faux
```diff
@@ -87,7 +87,23 @@ if [ -n "${options[xdhcp]}" ]; then
     fi
     ip addr show $intf_name
 
-    (while true; do ping -c 1 10.20.0.1; sleep 5; done) &
+    if [ -n "${options[opendns]}" ]; then
+        route add -net 172.17.0.0/16 $intf_name  # default docker ip range.
```
grafnu, line 91: I think this should be 10.20.0.0

docker/include/bin/start_faux
```diff
@@ -87,7 +87,23 @@ if [ -n "${options[xdhcp]}" ]; then
...
+        echo nameserver ${dhcp_dns#DNS=} > /etc/resolv.conf
+        # There is no deterministic way to find out the default gateway ip where running more than 1 faux device.
+        (
+            while true; do
+                if [ `route -n | grep UG | wc -l` -eq '1' ]; then 
+                    timeout 1 ping -c 1 `route -n | grep UG ` && continue
+                fi 
+                for IP in {1..12}
+                do
+                    timeout 1 ping -c 1 172.17.0.$IP  && route add default gw 172.17.0.$IP $intf_name && break
```
grafnu, line 102: 10.20.0.$IP?

docker/include/bin/start_faux
```diff
@@ -87,7 +87,23 @@ if [ -n "${options[xdhcp]}" ]; then
...
+                done
+            done
+        ) &
```
grafnu, line 105: does this handle the case where the GW address changes?   I think the address can keep changing... ?

docker/include/bin/start_faux
```diff
@@ -87,7 +87,23 @@ if [ -n "${options[xdhcp]}" ]; then
...
-    (while true; do ping -c 1 10.20.0.1; sleep 5; done) &
+    if [ -n "${options[opendns]}" ]; then
+        route add -net 172.17.0.0/16 $intf_name  # default docker ip range.
+        dhcp_dns=8.8.8.8
+        echo nameserver ${dhcp_dns#DNS=} > /etc/resolv.conf
+        # There is no deterministic way to find out the default gateway ip where running more than 1 faux device.
+        (
+            while true; do
+                if [ `route -n | grep UG | wc -l` -eq '1' ]; then 
+                    timeout 1 ping -c 1 `route -n | grep UG ` && continue
```
grafnu, line 98: why is timeout necessary?  ping should timeout automatically?

docker/include/bin/start_faux
henry54809: 172.17.0.0/16 is the external eth0 ip. For some reason the 10.20.X.X gateway IPs vary too broadly when gateways are restarted for bruteforce

docker/include/bin/start_faux
grafnu, line 100: i'm not sure this will work... the gateway IP address can get arbitrarily large since it's dynamically allocated. I just checked my run and it was up to 36. We might need to do something explicitly like write a value to a mapped file in faux

docker/include/bin/start_faux
henry54809: Yes so if ping to gateway stops responding, it will go through the bruteforce again

docker/include/bin/start_faux
henry54809: default is 4 seconds, so it takes too long to go through the list

docker/include/bin/start_faux
grafnu: does it keep adding default routes, or do they get removed somehow?

docker/include/bin/start_faux
henry54809: Fixed.

docker/include/bin/start_faux
henry54809: I forgot to remove them but they still worked fine. I think system will block off the dead one.

docker/include/bin/start_faux
grafnu: I think mine is configured differently (see below).  Should be fine as long as it works in CI and for Karthik... and for the qual lab operators. Can you maybe add a check in that will hard-fail if opendns is specified but the subnet is not 172.17.x.x?  I'm trying to avoid weird submarine bugs out there.  The reason the 10.20.x.x gateways vary so much is they're dynamically allocated and are chosen separately for each new instance of a gateway... so for an extended run they can get arbitrarily large.

```
~/daq$ docker exec daq-gw01 ip addr
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
1597: eth0@if1598: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default 
    link/ether 02:42:c0:a8:09:04 brd ff:ff:ff:ff:ff:ff link-netnsid 0
    inet 192.168.9.4/24 brd 192.168.9.255 scope global eth0
       valid_lft forever preferred_lft forever
1599: gw01-eth0@if1600: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default qlen 1000
    link/ether 6a:11:f8:b3:cb:10 brd ff:ff:ff:ff:ff:ff link-netnsid 0
    inet 10.20.0.71/16 brd 10.20.255.255 scope global gw01-eth0
       valid_lft forever preferred_lft forever
    inet 10.20.92.2/32 scope global gw01-eth0
       valid_lft forever preferred_lft forever
    inet 192.168.84.1/32 scope global gw01-eth0
       valid_lft forever preferred_lft forever
```

docker/include/bin/start_faux
grafnu: Most likely just takes the most recent since I don't think there's any way for the system to know which is dead... but either way as long as it works.

henry54809:
Updated. PTAL

PR #577, opened 2020-08-05
Minor cleanup

grafnu:


henry54809:
I think the badge on README also needs to be changed when workflow name changes.

PR #580, opened 2020-08-07
DHCP Long (10)

frgitdaq:


config/modules/all.conf
```diff
@@ -8,3 +8,6 @@ include subset/security/build.conf
 include subset/cloud/build.conf
 include subset/manual/build.conf
 include subset/network/build.conf
+
+# DHCP configuration.
```
grafnu, line 12: comment is redundant/unnecessary

What's the broader category this falls under?  EIther it should be under "baseline" if it's core to the system, or it should just be merged with host.conf, or we need to categorize it appropriately. 

My guess is that dhcp.conf should just be merged into host.conf -- but I see you explicitly pulled the ipaddr tests out from there... why?  Is it that there's expected to be more "dhcp tests" in the future?

daq/ipaddr_test.py
```diff
@@ -60,6 +62,15 @@ def _dhcp_port_toggle_test(self):
         self.host.connect_port(True)
         self._ip_callback = self._next_test
 
+    def _dhcp_lease_wait_test(self):
+        if not self.host.connect_port(False):
+            self.log('disconnect port not enabled')
+            return
+        lease = int(self.lease_time[:-1])
```
grafnu, line 69: need check that trailing character is actually 's' -- otherwise very confusing if somebody inadvertently enters "100" instead of "100s"... Need to make that an obvious error.

daq/ipaddr_test.py
```diff
@@ -60,6 +62,15 @@ def _dhcp_port_toggle_test(self):
         self.host.connect_port(True)
         self._ip_callback = self._next_test
 
+    def _dhcp_lease_wait_test(self):
+        if not self.host.connect_port(False):
+            self.log('disconnect port not enabled')
```
grafnu, line 67: what happens in this case.... is the error silently ignored?  Does it show up as skip? fail? pass? fatal error?

config/system/default.yaml
```diff
@@ -32,7 +32,7 @@ site_path: local/site/
 
 # DHCP lease time parameters
 initial_dhcp_lease_time: 120s
-dhcp_lease_time: 500s
+dhcp_lease_time: 120s
```
grafnu, line 35: this doesn't seem right... the ongoing lease time shouldn't be so short (not realistic behavior). What's the impetus behind this change?

resources/setups/baseline/module_config.json
```diff
@@ -1,10 +1,13 @@
 {
   "modules": {
     "ipaddr": {
-      "enabled": false,
+      "enabled": true,
       "timeout_sec": 900,
       "port_flap_timeout_sec": 20,
-      "dhcp_ranges": [{"start": "192.168.0.1", "end": "192.168.255.254", "prefix_length": 16}]
+      "dhcp_ranges": [{"start": "192.168.0.1", "end": "192.168.255.254", "prefix_length": 16},{"start": "10.0.0.1", "end": "10.255.255.254", "prefix_length": 8},{"start": "172.16.0.1", "end": "172.31.255.254", "prefix_length": 12}]
```
grafnu, line 7: can we format this a bit nicer so it makes a good example?  All the dhcp_ranges json on one line is very ugly!

testing/test_many.out
```diff
@@ -2,12 +2,12 @@ Running testing/test_many.sh
 Many Tests
 DAQ stress test
 Enough results: 1
-Enough DHCP timeouts: 1
+Enough DHCP timeouts: 0
```
grafnu, line 5: this seems wrong -- why are these changing and/or failing?

config/modules/all.conf
frgitdaq: This is to ensure ipaddr and DHCP runs after all other tests. Otherwise the modified network configuration [interferes with other tests (notably security.passwords)](https://github.com/faucetsdn/daq/runs/970603327). An alternative approach would be to take a second look at how ipaddr and the faux device behaves after the functions have finished (ideally returning to the original IP address and configuration).

daq/ipaddr_test.py
frgitdaq: @grafnu I was testing this behaviour. It seems as if when there is no 's' at the end of the lease time, DAQ fails when providing the initial DHCP lease. However, I will make this change.

config/system/default.yaml
frgitdaq: @grafnu Simply to reduce the overall time that DAQ takes to run. 500s is fine but to be most accurate with the DHCP lease time wait test, a wait (equivalent to the total DHCP lease time) should be imposed to ensure a complete DHCP lease exipiry occurs. Therefore, without this change, DAQ will take 500s on this single test.

daq/ipaddr_test.py
frgitdaq: @grafnu If this test is not successful in ipaddr.py, no wait time occurs or a port toggle resulting in no DHCP request. Within the DHCP test code, this will return a fail on faux interface 2 (as expected anyway) and also a fail on faux interface 1. If USI were to fail I do not believe there is a way for this to be detected in the DHCP test code due to the separation between docker tests and USI - so an ideal skip case would not be possible.

testing/test_many.out
frgitdaq: because there aren't any DHCP timeouts

frgitdaq:
PR put into draft until PR #579 is closed. To prevent redundant code review

resources/setups/baseline/module_config.json
frgitdaq: See commit c7cff772639e46112af4a7daff51ad0539e291af

config/modules/all.conf
grafnu: Update comment to say why, not what.  https://blog.codinghorror.com/code-tells-you-how-comments-tell-you-why/

daq/ipaddr_test.py
grafnu: My question is what happens after logging 'disconnect port is not enabled' occurs -- does it just log the error message then nothing else? What's the observable result of this condition?  The log file should explain why something happened, but it should be obvious by other behaviors that there was a problem in the first place.

config/system/default.yaml
grafnu: If this is just for one test then it should be applied only to that one test and not as a blanket global.  The 500s is there for a reason since it's more typical behavior. And this file is not only for testing.

testing/test_many.out
grafnu: why aren't there any DHCP timeouts?

noursaidi:
@frgitdaq I've had a review, have the following initial comments:

- There is a discontinuity between what the DHCP module does and the IPADDR module does, for example:
  - Sometimes the IPADDR module initiates, runs the port-toggle, and then there’s a little pause while the framework does other stuff/runs other tests. In the meanwhile, the device may renew its DHCP lease, which is sometimes misconstrued as a pass for the long lease test, even if the second “long lease” disconnect hasn’t happened yet. 
  - I had one instance where the the “long_lease” disconnect happened a second after the “port_toggle”, and there were no DHCP packets in the packet capture between the two tests. The "short DHCP" test ended up passing, but the "long DHCP" failed, which is opposite to what was expected. This only happened once though, not sure why. @henry54809 any thoughts?
- The test document says for long lease it should look for a DHCPREQUEST packet which is what you’re doing, I’m just wondering if it should also/instead look for a DHCPDISCOVER instead, because that’s the expected behavior when a lease expires. @pisuke thoughts?

frgitdaq:
> @frgitdaq I've had a review, have the following initial comments:
> 
> * There is a discontinuity between what the DHCP module does and the IPADDR module does, for example:
>   
>   * Sometimes the IPADDR module initiates, runs the port-toggle, and then there’s a little pause while the framework does other stuff/runs other tests. In the meanwhile, the device may renew its DHCP lease, which is sometimes misconstrued as a pass for the long lease test, even if the second “long lease” disconnect hasn’t happened yet.
>   * I had one instance where the the “long_lease” disconnect happened a second after the “port_toggle”, and there were no DHCP packets in the packet capture between the two tests. The "short DHCP" test ended up passing, but the "long DHCP" failed, which is opposite to what was expected. This only happened once though, not sure why. @henry54809 any thoughts?
> * The test document says for long lease it should look for a DHCPREQUEST packet which is what you’re doing, I’m just wondering if it should also/instead look for a DHCPDISCOVER instead, because that’s the expected behavior when a lease expires. @pisuke thoughts?

@noursaidi Totally agree with your observations. But unfortunately, this is something which is out of the control of the actual tests that I'm developing. As the google doc suggests, I'm just expected to analyse the pcap file to determine the DHCPREQUEST packet. 

If there is suggestion to change the way test is implemented then sure, I donot have any problem to ammend the test to include the DHCPDISCOVER as well.

Will wait for the comments. 
@spectre-storm adding you to the thread to make you aware of Nour's comments and suggested change in the implementation.

grafnu:
Given the asynchronous behaviour of the system (the DHCP mechanism is
largely completely out of control of the tests execution), some amount of
flaky / indeterminate behaviour is possible. I don't think it'll be
tractable to make some of this "work 100% of the time" -- it will, as you
cite, occasionally do the unexpected.  The solution to this is what we
generally talk about as "soak testing" -- which really mans we run the
tests (e.g.) 10 times and then see if it passes >80% of the time (or
something like that).

In short, I don't think some indeterminate behavior is a problem as long as
*most* of the time it's consistent.  We will need to catch this at some
other layer in the system, not at the level of a test/module.

So, if checking for DHCPDISCOVER does the right thing then all good -- but
if not for whatever reason -- I think it's OK and we just flag it as such
and resolve with the soak testing.  There are other instances of things
like this in the system so it's not the only situation where this arises.

frgitdaq:
@henry54809 There did not seem to be any framework implementation for this test. I have modified the ipaddr module to what I was expecting to see though it interferes with the integration testing. Am I missing something in the framework changes?

daq/test_modules/ipaddr_module.py
```diff
@@ -85,6 +86,15 @@ def _dhcp_port_toggle_test(self):
         self.host.connect_port(True)
         self._ip_callback = self._next_test
 
+    def _dhcp_long_wait_test(self):
+        self._set_timeout()
+        if not self.host.connect_port(False):
+            self._logger('disconnect port not enabled')
+            return
+        time.sleep(self._get_lease_time())
```
henry54809, line 94: This wasn’t part of the 3 extended DHCP test when I was implementing these. I thought we agreed on this test was fundamentally no different from the port_toggle considering  when you disconnect and reconnect the port, regardless of how much wait time, the device is expected to request DHCP immediately. 
time.sleep is also a blocking call so waiting for an extended period of time in python shouldn’t be done this way.

frgitdaq: @spectre-storm @pisuke Could you please comment on this? And if this is the case, how to go about implementing the dhcp_long test?

frgitdaq:
PR closed as test no longer required.

PR #582, opened 2020-08-09
refactoring for vlan trigger

henry54809:


cmd/exrun
```diff
@@ -74,6 +74,12 @@ rm -rf $INSTDIR/faucet* $cleanup_file
 
 docker ps > /dev/null 2>&1 || service docker start
 
+bridges=`ovs-vsctl list-br`
+for bridge in $bridges; do
+    echo Cleaning bridge $bridge...
+    timeout 10 ovs-vsctl --if-exists del-br $bridge || true
```
grafnu, line 80: add comment explaining why this is run twice... ?  IIRC this is some weird bug in ovs, right?

daq/host.py
```diff
@@ -817,11 +821,16 @@ def _initialize_config(self):
...
         self._record_result(None, config=self._make_config_bundle())
 
     def _release_config(self):
         self._gcp.release_config(self._DEVICE_PATH % self.target_mac)
-        self._gcp.release_config(self._CONTROL_PATH % self.target_port)
+        if self.target_port:
+            self._gcp.release_config(self._CONTROL_PATH % self.target_port)
+
+    def __repr__(self):
+        return self.target_mac + " on port %d" % self.target_port if self.target_port else ""
```
grafnu, line 836: add parens to clarify expression grouping (improves readability)

daq/docker_test.py
```diff
@@ -18,13 +20,15 @@ class DockerTest:
...
-    def __init__(self, host, target_port, tmpdir, test_name, module_config):
-        self.target_port = target_port
+    def __init__(self, host, tmpdir, test_name, module_config):
+        self.device = host.device
         self.tmpdir = tmpdir
         self.test_name = test_name
         self.runner = host.runner
-        self.host_name = '%s%02d' % (test_name, self.target_port)
+        # Host name can't be more than 15 characters.
+        # The name is also used to create interface in mininet
```
grafnu, line 29: Are these two statements logically connected?  If so, add a "because" in there to link them, otherwise the second commend doesn't have much context.

daq/docker_test.py
```diff
@@ -18,13 +20,15 @@ class DockerTest:
...
+        self.host_name = '%s_%s' % (test_name[:6], ''.join(
+            random.choice(string.ascii_letters) for i in range(3)))
```
grafnu, line 31: what is this doing?  Maybe extract to named function to make it more clear? Or create named intermediate variable to give a hint as to what the randomness is for.

daq/docker_test.py
```diff
@@ -108,26 +112,26 @@ def _check_image(self):
 
     def terminate(self):
         """Forcibly terminate this container"""
-        LOGGER.info("Target port %d test %s terminating", self.target_port, self.test_name)
+        LOGGER.info("%s terminating", self)
```
grafnu, line 115: Why is it sometimes "Target device %s" % self.device, and other times just "%s" % self?  Line 104 vs 115... I would expect them to be the same?

testing/test_aux.out
```diff
@@ -69,9 +69,9 @@ RESULT skip connection.network.ntp_support No NTP packets received.
 RESULT skip connection.network.ntp_update Not enough NTP packets received.
 RESULT pass connection.mac_oui Manufacturer: Google found for address 3c:5a:b4:1e:8f:0a
 dhcp requests 1 1 1 1
-01: []
-02: ['02:ping:TimeoutError']
-03: []
+3c:5a:b4:1e:8f:0a: []
```
grafnu, line 72: this should be the form without :'s

daq/runner.py
```diff
@@ -22,21 +22,90 @@
 LOGGER = logger.get_logger('runner')
 
 
-class PortInfo:
+class PortInfo():
```
grafnu, line 25: leave off ()

daq/runner.py
```diff
@@ -22,21 +22,90 @@
...
-    flapping_start = 0
-    mac = None
-    host = None
-    gateway = None
-
+    flapping_start = None
+    port_no = None
+
+
+class IpInfo():
```
grafnu, line 32: no ()

henry54809:
PTAL.

PR #584, opened 2020-08-10
Private Address (6)

frgitdaq:


config/modules/all.conf
```diff
@@ -8,3 +8,6 @@ include subset/security/build.conf
 include subset/cloud/build.conf
 include subset/manual/build.conf
 include subset/network/build.conf
+
```
grafnu, line 11: these changes are redundant with the other PR?  I don't want to review things twice, so if there is significant overlap then we'll need to serialize the reviews.  You could also do a PR against your other branch, which would then just show the difference between them allowing for a review of just the deltas.

frgitdaq: Will put the status of this PR into draft and revert it to ready for review once PRs for tests 9 and 10 have been merged. It was agreed that each test would be a PR merge onto master. 
Apologies for the redundant changes across PRs.

frgitdaq:
PR put into draft until PRs #579 and #580 are closed. To prevent redundant code review

PR #588, opened 2020-08-12
Feature/vlan trigger

henry54809:


config/system/default.yaml
```diff
@@ -41,6 +41,8 @@ finish_hook: bin/dump_network
 # topology hook: executed when device topology changes
 topology_hook: bin/dump_network
 
+test_trigger_type: PORT
```
grafnu, line 44: Trying to avoid 'test' as an adjective since it's very overloaded... also, technically I don't think this is triggering a 'test", as such... run_trigger_type?  sequence_trigger_type?  runner_trigger_type?  How would you describe in one word what it's triggering?

henry54809: Updated name. PTAL

PR #595, opened 2020-08-14
Update NTP support test fix #622

### LINKED ISSUE #622: NTP Update Test Validity
I am not sure the current requirements behind connection.network.ntp_update are actually valid.  Since this test was introduced, it always fails even on known good devices that should be passing this test.  I have concerns that either the implementation and/or the qualification definitions around this test are not adequate in generating pass results.

frgitdaq:
Use monitor.pcap rather than startup.pcap when detecting the NTP version.


subset/network/ntp_tests.py
```diff
@@ -63,9 +63,13 @@ def ntp_payload(packet):
 
 
 def test_ntp_support():
-    capture = rdpcap(startup_pcap_file)
-    if len(capture) > 0:
-        version = ntp_client_version(capture)
+    startup_capture = rdpcap(startup_pcap_file)
```
grafnu, line 66: you shouldn't actually use startup_pcap_file at all, unless you are explicitly trying to track something during the DHCP exchange.  If there is no monitor_pcap then just return false.

noursaidi:
@pisuke merge conflict resolved, I think this can be merged

PR #600, opened 2020-08-18
Refactor base test module and use stable test set id

grafnu:


daq/runner.py
```diff
@@ -55,19 +56,33 @@ class Devices:
     """Container for all devices"""
     def __init__(self):
         self._devices = {}
+        self._set_ids = set()
 
-    def new_device(self, mac):
+    def new_device(self, mac, port_info):
```
henry54809, line 61: I think it's cleaner to use def new_device(self, mac, port_info=None):
so you don't have to pass None when it's not needed.

grafnu: agreed -- python isn't my native language so still use old-style constructs sometimes!

PR #601, opened 2020-08-18
Feature/convert switchtests

henry54809:


daq/host.py
```diff
@@ -699,6 +699,12 @@ def _get_switch_config(self):
             'password': self.switch_setup.get('password')
         }
 
+    def _get_usi_config(self):
+        return {
```
grafnu, line 703: how is this different than just returning self._usi_config?  The naming of things suggest one of two things 1) there's a comment required b/c the "why" is not at all clear, 2) this code is highly redundant and can be greatly simplified.

subset/switches/src/main/java/switchtest/Main.java
```diff
@@ -1,50 +1,46 @@
...
+   * Switch test runner.
+   * @param args args
+   */
+  public static void main(String[] args) {
 
-    if (args.length < 4) {
+    if (args.length < 5) {
       throw new IllegalArgumentException(
-          "Expected ipAddress && port && supportPOE && switchModel as arguments");
+          "usiUrl, rpcTimeoutSec switchIpAddress, port, supportPOE, switchModel [username] [password] as arguments");
```
grafnu, line 18: something about this string is not right. Lost "expected"?  Some things have commas others don't?

subset/switches/src/main/java/switchtest/SwitchTest.java
```diff
@@ -0,0 +1,156 @@
...
+  protected int rpcTimeoutSec;
+
+  /**
+   * Generic switch test.
+   * @param channel GRPC channel
+   * @param rpcTimeoutSec Timeout in seconds for rpc calls
+   * @param deviceConfigPoeEnabled poe config from module_config
+   * @param debug print debug output
+   */
+  public SwitchTest(Channel channel, int rpcTimeoutSec, boolean deviceConfigPoeEnabled,
```
grafnu, line 33: nb: something is still wrong with your settings or the linter. This should just be hanging 4 indent (as per Google styleguide).  Not a big deal so don't need to fix it but just be aware.

subset/switches/src/main/java/switchtest/SwitchTest.java
```diff
@@ -0,0 +1,156 @@
...
+      }
+    }
+    return "RESULT fail connection.port_speed Cannot detect current speed\n";
+  }
+
+  protected String testDuplex(InterfaceResponse interfaceResponse) {
+    String duplex = interfaceResponse.getDuplex();
+    if (duplex != null) {
+      if (duplex.equals("full")) {
+        return "RESULT pass connection.port_duplex Full duplex mode detected\n";
```
grafnu, line 66: These strings get ugly after a while... I'd consider extracting to a function or format. E.g. you could have result_string(Boolean pass, String test_name, String detail) and then just say:
return result_string(true, "port_duplex", "Full duplex mode detected")

Generally it's bad form to have a lot of embedded constants, especially a lot of embedded strings.

subset/switches/src/main/java/switchtest/SwitchTest.java
```diff
@@ -0,0 +1,156 @@
...
+    try {
+      if (debug) {
+        System.out.println("report:" + loginReport);
+      }
+
+      String[] directory = reportFilename.split("/");
+
+      File dir = new File(directory[directory.length - 2]);
+      if (!dir.exists()) {
+        dir.mkdirs();
```
grafnu, line 130: check for exists() is redundant for mkdirs() -- it will silently do nothing if the directory already exists. (at least I'm 99% sure)

subset/switches/src/main/java/switchtest/SwitchTest.java
henry54809: This passes the linter on my local machine and on stickler. It can only mean the checkstyle file doesn't include this check.

daq/host.py
henry54809: I wanted to make sure the keys were always there since usi_config can be empty but it looks like jq doesn't fail if it's indexing a key that doesn't exist.

henry54809:
PTAL

subset/switches/readme.md
```diff
@@ -68,6 +68,10 @@ Example of all necessary parameters in the system.conf related to physical switc
 	#Define the password for the switch. This parameter is optional.
 	switch_setup.password=switch_p@55
 
+    #Define the usi url using your docker0's ip and port 5000, and re-run bin/setup_base if you're upgrading from versions before 1.9.0
```
grafnu, line 71: spaces after #s

PR #602, opened 2020-08-19
Auto rewrite usi_set.url 

henry54809:


cmd/exrun
```diff
@@ -118,7 +118,12 @@ else
 fi
 
 docker rm -f daq-usi || true
-autostart cmd/usi
+uri_url_override=""
```
grafnu, line 121: put at top of file along with over nulls and defaults

subset/switches/readme.md
```diff
@@ -70,7 +70,7 @@ Example of all necessary parameters in the system.conf related to physical switc
 
     # Define the usi url using your docker0's ip and port 5000, and re-run bin/setup_base if you're upgrading from versions before 1.9.0
     # Make sure docker's ip range doesn't conflict with that of the switch. Default docker0 ip is 172.17.0.1 
-    usi.url=172.17.0.1:5000
+    usi_setup.url=172.17.0.1:5000
```
grafnu, line 73: Is this step still necessary if it auto-defaults?   Maybe say only do this if it's not the default local docker0 interface?

henry54809:
PTAL

PR #606, opened 2020-08-20
USI debug messages 

henry54809:


usi/src/main/java/daq/usi/UsiImpl.java
```diff
@@ -14,23 +14,25 @@
 
 public class UsiImpl extends USIServiceGrpc.USIServiceImplBase {
   private final Map<String, SwitchController> switchControllers;
+  private boolean debug;
```
grafnu, line 17: nit: final

usi/src/main/java/daq/usi/UsiServer.java
```diff
@@ -48,12 +51,13 @@ private void blockUntilShutdown() throws InterruptedException {
 
   /**
    * Main method.
-   * @param args not used.
+   * @param args "debug".
    * @throws Exception Maybe a refactor is needed to throw more specific exceptions.
    */
   public static void main(String[] args) throws Exception {
+    final boolean debug = args.length > 0 ? args[0].trim().equals("debug") : false;
```
grafnu, line 58: trim is unnecessary

PR #615, opened 2020-08-23
USI docker gcplogs config

henry54809:
Can you please test it out on your machine first? The docs on gcplogs driver isn't very clear on the configs needed. 
It requires a restart of docker service for the config to be reloaded, so it may take some time to run cmd/usi.


henry54809:
PTAL

PR #616, opened 2020-08-24
Use daq_run_id for gcp combine_results

grafnu:


bin/python/combine_reports.py
```diff
@@ -132,17 +134,23 @@ def _convert_iso(timestamp):
...
     [local/system.yaml]
     device=xx:xx:xx:xx:xx:xx
-    from_time='YYYY-MM-DDThh:mm:ss'
-    to_time='YYYY-MM-DDThh:mm:ss'
+    [from_time='YYYY-MM-DDThh:mm:ss']
+    [to_time='YYYY-MM-DDThh:mm:ss']
+    [daq_run_id=<nonce>]
     [count=N]
-    [from_gcp='true']
+    [from_gcp='true/false']
```
henry54809, line 149: any string passed here (including false) will be considered true under CONFIG.get('from_gcp'). We can either check if the string is true / false, or revert the doc to only indicate true as an option.

grafnu: Got it... actually had that fixed but sitting in an unpushed commit on my machine!

PR #619, opened 2020-08-27
Native tests -- Draft. **This will break your host machine**

henry54809:
**Inheritance**
HostModule <- External Test <- Docker Test 
                                                       - Creates a Docker Host
                                                  <- Native Test    
                                                        - Creates a Native Host

Some of the folder mappings aren't right, but the plan is to create another directory under inst/run**/nodes/*/root where everything in the test folder will be linked to. That folder will be the root folder for the test, and is available as a env_var TEST_ROOT for script to access.


daq/test_modules/external_test.py
```diff
@@ -0,0 +1,153 @@
+"""Module for running docker-container tests"""
+
```
grafnu, line 2: should be external_module -- should start using the new nomenclature.

daq/test_modules/docker_test.py
```diff
@@ -0,0 +1,49 @@
...
+import logger
+from clib import docker_host
+
+from .external_test import ExternalTest
+
+
+LOGGER = logger.get_logger('docker')
+
+
+class DockerTest(ExternalTest):
```
grafnu, line 13: probably change this to DockerModule while we're moving things around.

daq/test_modules/native_host.py
```diff
@@ -0,0 +1,100 @@
+"""A docker-based mininet host"""
```
grafnu, line 1: not docker

daq/test_modules/native_test.py
```diff
@@ -0,0 +1,44 @@
...
+import logger
+
+from .native_host import make_native_host
+from .external_test import ExternalTest
+
+
+LOGGER = logger.get_logger('docker')
+
+
+class NativeTest(ExternalTest):
```
grafnu, line 13: NativeModule not Test

daq/host.py
```diff
@@ -79,6 +78,11 @@ class ConnectedHost:
     _AUX_DIR = "aux/"
     _CONFIG_DIR = "config/"
     _TIMEOUT_EXCEPTION = TimeoutError('Timeout expired')
+    _NATIVE_TESTS_MAPPING = {
```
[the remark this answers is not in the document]
henry54809: yeah I was planning on extracting this test mapping out when I deal with the dependency issue in nmap.

daq/test_modules/external_test.py
henry54809: should the ipAddrTest be a module as well?

grafnu:
yes...

PR #625, opened 2020-09-03
Feature/native tests

henry54809:


subset/ping/ping.daqmodule
```diff
@@ -0,0 +1,4 @@
+{
+    "name": "ping",
+    "startupScript": "test_ping"
```
grafnu, line 3: should be startup_script following other naming conventions

testing/test_preamble.sh
```diff
@@ -67,7 +67,9 @@ function redact {
         -e 's/Not shown: .* ports//' \
         -e 's/[ \t]*$//' \
         -e 's/\t/ /g' \
-        -e 's/([0-9]{1,3}\.){3}[0-9]{1,3}/X.X.X.X/'
+        -e 's/([0-9]{1,3}\.){3}[0-9]{1,3}/X.X.X.X/' \
+        -e 's/-oG .*\/tmp/-oG \/tmp/' \
+        -e 's/# Nmap [0-9]{1,4}\.[0-9]{1,4}/\# Nmap/'
```
grafnu, line 72: replace version with marker (XXX) not nothing

testing/test_preamble.sh
```diff
@@ -67,7 +67,9 @@ function redact {
         -e 's/Not shown: .* ports//' \
         -e 's/[ \t]*$//' \
         -e 's/\t/ /g' \
-        -e 's/([0-9]{1,3}\.){3}[0-9]{1,3}/X.X.X.X/'
+        -e 's/([0-9]{1,3}\.){3}[0-9]{1,3}/X.X.X.X/' \
+        -e 's/-oG .*\/tmp/-oG \/tmp/' \
```
grafnu, line 71: replace removed things with marker, not nothing

testing/test_modules.sh
```diff
@@ -10,14 +10,9 @@ cp config/system/all.conf local/system.conf
 TEST_LIST=/tmp/module_tests.txt
 
 cat > $TEST_LIST <<EOF
-ping
```
grafnu, line 13: why remove all these?

testing/test_utils.sh
```diff
@@ -46,14 +46,14 @@ function check_setup {
     cat >> $cmd_file <<EOF
 sleep 30
 function test_bacnet {
-    bacnet_base="tcpdump -en -r /scans/test_ping.pcap port 47808"
+    bacnet_base="tcpdump -en -r \$TEST_ROOT/scans/test_ping.pcap port 47808"
```
grafnu, line 49: cleaner to use single quotes around entire string rather than escape each $

testing/test_modules.sh
henry54809: test_modules only work with docker containers setups; I don't know if it's worth the investment to rewrite another one to work with 2 native tests.

daq/test_modules/native_module.py
```diff
@@ -0,0 +1,45 @@
...
+        env_vars = super()._get_env_vars(params)
+        env_vars.append(('TEST_ROOT', self.basedir))
+        return env_vars
+
+    def _get_vol_maps(self, params):
+        vol_maps = super()._get_vol_maps(params)
+
+        # Common testing tools
+        vol_maps.append((os.path.abspath('bin/retry_cmd'), '/bin/retry_cmd'))
+        vol_maps.append((os.path.abspath('resources/setups/baseline/module_manifest.json'),
```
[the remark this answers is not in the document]
henry54809: Why not remove all module_manifests and store them directly in module_configs.

daq/test_modules/native_module.py
```diff
@@ -0,0 +1,45 @@
...
+                         os.path.join(self.basedir, 'module_manifest.json')))
+        vol_maps.append((os.path.abspath('docker/include/utils/reporting.sh'),
```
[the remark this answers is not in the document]
henry54809: This, to me, is similar to bin/retry_cmd that I think it should be hard coded. This is also in sync with the base docker container : https://github.com/faucetsdn/daq/blob/master/docker/modules/Dockerfile.aardvark

subset/ping/ping.daqmodule
henry54809: I followed the json naming convention here but it's not ideal, I can convert to snake case.

grafnu: understood... JSON ends up being all over the place and really has nothing to do with JS anymore :-).  Generally I've found that snake_case is better than cAmel case and Java/JS just got it wrong.  Anyway -- my general guidance is if possible go with snake_case even if it's "wrong" (e.g. sometimes it shows up in Java code too...)

testing/test_modules.sh
grafnu: fair enough. As long as it's all being tested somewhere (which it is still in the integration tests) then we're all good. I agree that doing more work in this direction doesn't make sense until we have the wider team trying to do native test development.

testing/test_utils.sh
henry54809: Couldn’t get it working with single quotes. They’re evaluated too early if they’re wrapped in single quotes.  PTAL.

daq/test_modules/native_host.py
```diff
@@ -27,7 +27,8 @@ def __init__(self, name, basedir=None, tmpdir=None, env_vars=None, vol_maps=None
         self.vol_maps.append((os.path.abspath(os.path.join(self.tmpdir, 'tmp')),
                               os.path.join(self.basedir, 'tmp')))
         self.name = name
-        self.startup_script = os.path.join(self.basedir, startup_script)
+        self.startup_script = startup_script if startup_script.startswith('/') \
```
grafnu, line 30: avoid continuation, use temporary variable instead

resources/setups/baseline/module_config.json
```diff
@@ -14,7 +14,21 @@
       "enabled": true
     },
     "switch": {
-      "enabled": true
+      "enabled": true,
```
grafnu, line 17: what's this change for?  I understand that it's inlining the module_manifest.json file, which might be a good thing -- I'm just trying to understand the motivation..

henry54809: You mentioned earlier (https://github.com/faucetsdn/daq/pull/625#discussion_r486493462 ) that module_manifest mapping shouldn't use hardcoded path, so I thought why not merge module_manifest into module_config. I think the end result is a lot cleaner.

grafnu: There is a semantic difference between them -- one is configuration that describes how a particular instance of something is, while the other is basic information about the test (but is conceptually immutable, unlike configuration).  It might be cleaner form a "it's just one file" perspective but it seems to me it's breaking abstractions.

daq/test_modules/native_module.py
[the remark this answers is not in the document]
grafnu: as mentioned in the other comment, a manifest is conceptually different than config

resources/setups/baseline/module_config.json
henry54809: I think the manifests should go in the module_config as they're really the configurations for the descriptions of tests.

grafnu:
I don't think they're the same thing.

daq/test_modules/native_module.py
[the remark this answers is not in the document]
henry54809: I really wish this could be discussed 5 days ago. Based on the comments, it looked like you were fine with this when I took a lot time to make all the changes. 
Base on the conceptually differences described, the manifest should really be into the new .daqmodule file; however this is also another big change.

grafnu:
Which comments are you referring to? The comment you linked was "don't
hardcode the path" -- the change should have been as simple as making the
path a configurable parameter. I took a look at the comment again and I
think you may have misunderstood -- it wasn't "fold everything in as a
config parameter" but rather "fold the path in as a config parameter".

Keeping parity with the docker modules, it would still be in a
module_manifest.json file just like everything else (the '.daqmodule' file
isn't currently universally available, e.g. for docker tests). One point of
confusion might be that the referenced module_manifest.json file is
currently in resources/setups/baseline/ and that's the wrong place for it,
since it's not semantically setup with a resource/setup, and rather should
have been somewhere more intrinsic with the code that is using it. But
that's not a structural change so didn't necessarily need to address that
now.

henry54809:
I suggested merging the 2 files under the linked comment but there were no objections so I interpreted that as ok. 
When I made the hardcoded string, I assumed this file(resources/setups/baseline/module_manifest.json) is actually the baseline module_manifest file(similar to baseline module_config, that's not subject to change), but it turns out to be not the case. Where is the configurable parameter going to be stored... in module_config? 
On a separate note, I wanted to make .daqmodule file universally available so we can remove the all .conf files. All modules, regardless of type have all the build / run information stored in there and module_config will be the only place for turning test modules on and off. I made the new .daqmodule files with this in mind but that'a discussion for later.

henry54809:
PTAL

grafnu (approved):

PR #626, opened 2020-09-05
Tls updates

jhughesbiot:
This PR should be considered a working PR to resolve two issues:
1. Get some initial feedback
2. Determine build failure differences between local validation and CI
  a. test_modules  - runs successfully on my local setup but completely fails the tls module in the CI
  b. test_aux - locally runs with no diffs but has diffs once pushed and run in CI


noursaidi:
@jhughesbiot I'm still reviewing, but I'm wondering if the TLS certificate test be a separate test to the server test as per the proposed test description, or has there been a change and the document now needs to be updated?

I don't think ECDH/ECDSA  exist as cipher suites for TLS1.3, so the reference to these in the TLS1.3 test should be removed.

jhughesbiot:
@noursaidi 
There has been some ongoing discussions around the test implementation and test document needs to be updated to match the change in requirements. I just updated the readme to reflect the ongoing discussions around these tests.  I will forward you the email threads where these discussions took place for review but are pretty well summarized in the updated readme.

jhughesbiot:
@noursaidi @grafnu @pisuke 
Pinging this thread to keep it moving

bin/test_module
```diff
@@ -56,15 +56,33 @@ docker_args+=" -v $module_dir/scans:/scans"
...
+
 mkdir -p $config_dir
 docker_args+=" -v $config_dir:/config/inst"
 docker_args+=" -v $config_dir:/config/port"
 docker_args+=" -v $config_dir:/config/device"
 docker_args+=" -v $config_dir:/config/type"
 
+# If testing the tls module, we need a few
+# additional things setup
+if [[ "$module" == "tls" ]]; then
```
grafnu, line 76: this doesn't need to be conditional.  Does it hurt anything if the .pem file is always provided?

bin/test_module
```diff
@@ -102,6 +120,14 @@ if [ -n "$failed" ]; then
     false
 fi
 
+#now interrupt the process.  get its PID:
+echo Stopping module pcap...
+pid=$(ps -e | pgrep tcpdump)
```
grafnu, line 125: use $! from when after the command is executed, rather than trying to extract the PID.

docker/include/bin/start_faux
```diff
@@ -215,10 +215,14 @@ if [ -n "${options[tls]}" ]; then
     echo Starting tls server on port 443 https.
     mv /root/nginx/nginxpass.conf /etc/nginx/nginx.conf
     service nginx start
+    python tlsfaux/client.py 1.2 google.com &
+    python tlsfaux/client.py 1.3 google.com &
 elif [ -n "${options[expiredtls]}" ]; then
     echo Starting expired tls server on port 443 https.
     mv /root/nginx/nginxfail.conf /etc/nginx/nginx.conf
     service nginx start
+    python tlsfaux/client.py 1.2 not-a-good-url.google.com/ &
```
grafnu, line 224: should URLs be the same?  (one has /, the other doesn't)

bin/test_module
```diff
@@ -56,15 +56,33 @@ docker_args+=" -v $module_dir/scans:/scans"
...
 docker_args+=" -v $config_dir:/config/port"
 docker_args+=" -v $config_dir:/config/device"
 docker_args+=" -v $config_dir:/config/type"
 
+# If testing the tls module, we need a few
+# additional things setup
+if [[ "$module" == "tls" ]]; then
+  #CA file is neededto validate the certs properly so we need to
+  #extract it from the faux device
+  CA_FILE=$(docker exec daq-$faux_intf cat ./tlsfaux/certs/CA_Faux.pem)
```
grafnu, line 79: we shouldn't be extracting this from the docker image... If it's needed elsewhere then we should create it during the build step and then copy it into the faux file and/or this place.  Or, in this case (testing) we can likely just generate the keys and then have them in the source tree because it's only used internally for testing.

docker/include/security/tlsfaux/client.py
```diff
@@ -0,0 +1,37 @@
+import socket, ssl, pprint, absolute_filepath, sys, time
```
grafnu, line 1: looots of stickler comments so I can't really look at the contents of the file...

bin/test_module
jhughesbiot: It might start causing failures if it's not checked because it would never resolve anything unless it was the tls module.  This is due to the nature of the start_faux only calling generate_certs.py when the tls module is detected.  The same module detection method was used here to keep that specific scenario in sync.

subset/security/test_tls
```diff
@@ -6,11 +6,26 @@ LOCAL_REPORT=tmp/report.txt
...
 if [ -n "$TARGET_IP" ]; then
     echo Collecting TLS cert from target address %% $TARGET_IP > $MONO_LOG
 
-    java -jar tlstest/build/libs/tlstest.jar $TARGET_IP
+    export CA_FILE="None"
+    export CA_URL="None"
+    if [ -f "$MODULE_CONFIG" ]; then
+      echo "Module Config present"
+      CA_FILE=`jq -r .modules.security.tls.ca_file $MODULE_CONFIG`
+      CA_URL=`jq -r .modules.security.tls.ca_url $MODULE_CONFIG`
```
noursaidi, line 20: I think these should be modules.tls.ca_file

subset/security/readme.md
```diff
@@ -1,48 +1,54 @@
...
+### Testing procedure for server testing
 After establishing connections to devices, the test will proceed to validate the available certificates with the following criteria:
  1. The certificate is in the x509 format
- 2. The public key length is at least 2048.  Currently handles both RSA and DSA public key formats.
- 3. The certificate is not expired and active for the current date.
-The cipher suite used is also checked but does not currently affect the outcome of the results.  Currently the expected cipher suites are ECDH and ECSA.  If these are not present, a warning message will be logged in the activate.log of the switch node.
+ 2. The public key length is at the correct bit length
+    - If public key is RSA format - Key length must be at least 2048
+    - If public key is EC format - key length must be at least 224
+ 3. The certificate is not expired and active for the current date
```
noursaidi, line 12: Doesn't mention the signed certificate requirement in the read me

subset/security/readme.md
```diff
@@ -1,48 +1,54 @@
...
+ 4. The cipher suite is valid:(only for TLS < 1.3)
+    - If RSA public key format - No cipher check is required
+    - If EC pubic key format - Check if cipher suites ECDH and ECSA are present
 
- 
-### Note for test developers 
-The functional test code is included in the `tlstest/src/main/java` folder.
+### Testing procedure for client testing
```
noursaidi, line 17: Read me should give guidance for configuration (e.g. directory to the certificate file and what that certificate file is)

bin/test_module
jhughesbiot: Not sure what is being asked here.  Extraction of PID highlighted but indicating changing the command in some manner section to remove this necessity.

grafnu: https://stackoverflow.com/questions/1908610/how-to-get-process-id-of-background-process
You want to have `pid=$!` (at the appropriate point in the file), rather than greping through ps

bin/test_module
grafnu: fair enough. I think the bigger point is the one about how we shouldn't be extracting this from the faux device, I think.

bin/test_module
jhughesbiot: The certs are generated inside the faux container as it's started and are not available outside of that container.  If the cert generation is brought higher in the stack outside of the container start process, then that would require a significant amount of work and testing to verify all current tls parameters handle this new structure.  If necessary then we need to start that conversation of how that needs to look.

bin/test_module
jhughesbiot: If we move the cert generation to be available outside of the faux device, then we can update this at the same time.  I've responded on that topic so we can resolve that question there and then double back to this one after that's figured out.

bin/test_module
jhughesbiot: Got it, will implement.

bin/test_module
grafnu: I've had a think through this and I agree that extracting from the container is the lesser of all evils. I was trying out a refactoring and the problem quickly becomes one of distribution. Still doesn't make me happy, but less unhappy than alternatives.

jhughesbiot: Agreed. Maybe an update for a later date since it's a pretty tall order.

docker/include/security/tlsfaux/client.py
jhughesbiot: I've fixed most of these but there's still an issue with using print statements.  I can remove just not sure if this is an actual issue or just some default setting that should be updated on the stickler settings.

grafnu:
From reading around that seems to be a bug in pylint... Can you try adding
E1601 to .pylintrc (it's already there at the top-level) and see if it
makes those stickler warnings go away?

grafnu:
If that doesn't work then we can just ignore those for now and deal with it
later if necessary...

subset/security/readme.md
jhughesbiot: Updated, let me know if you think it needs more clarity.

subset/security/readme.md
jhughesbiot: Updated, let me know if you think it needs more clarity.

jhughesbiot:
@grafnu 
E1601 added to .pylintrc but seems to have no effect on the messages.

grafnu (approved):
Looks good to me at this point. I'm annoyed at Stickler but that has nothing to do with you or the PR.

grafnu:
Ok -- https://github.com/faucetsdn/daq/pull/637 seems to be the way to fix the stickler issues with print()

noursaidi:
When a server doesn't support a protocol, the test correctly fails but the error message says "_Certificate public key length too short. Certificate could not be validated. Certificate has not been signed by a CA_". Could this just say "Protocol not supported" or something to that degree?

Additionally, when a CA file is specified but it doesn't actually exist, the test fails and the report summary is "_Certificate not been signed by a CA_". I think it should still fail (?), but the message should highlight that CA specified was not found

jhughesbiot:
The server messaging reflects the specific failure points of the test.  There could exist a state where only some of the items are failing and the others are passing so the code is attempting to report all the failure conditions instead of a generic result of fail with little information to the why.

You can see the CA file resolving error in the activate.log or rather, you won't see the messaging around confirming the file was resolved and validation actually took place with the results.  I've added additional messaging to the activate.log when this scenario happens.

jhughesbiot:
@grafnu 
Updated stickler with that modification and it seems to have done the trick.

I think there's still a primary set of issues that hasn't been looked at before this can be accepted which I brought up initially in this PR.

Determine build failure differences between local validation and CI
a. test_modules - runs successfully on my local setup but completely fails the tls module in the CI
b. test_aux - locally runs with no diffs but has diffs once pushed and run in CI

Looks like the comments/feedback and updates have been mostly dealt with but the difference of local to CI builds is going to cause issues if it's accepted before we get those resolved.

grafnu:
Got it -- I'm replicating your branch locally and will try it out. I see
the failure in the actions logs, so I at least know what to look for.

grafnu:
Can you add "aardvark2" to the list of targets in testing/test_modules.sh https://github.com/faucetsdn/daq/blob/master/testing/test_modules.sh#L26

It looks like that module is built locally for you, but not in the CI. That change should fix that.

However, it still has a different error then -- which is something about the wrong # of arguments for a TLS client or something... but let's fix one thing at a time!

grafnu:
Re: test_aux -- this is the error from the logs. In order to diagnose further you'd have to put more debugging in there and compare with your local run.  Easiest thing to do would be to add the -x argument to the first line of test_tls.

```
2020-09-14T17:04:33.0217026Z ==> inst/run-9a02571e8f01/nodes/tls01/activate.log <==
2020-09-14T17:04:33.0217470Z Module Config present
2020-09-14T17:04:33.0219504Z ##[error]Exception in thread "main" java.lang.IllegalArgumentException: Expected 3 args, only 2 detected.
2020-09-14T17:04:33.0222122Z 	at Main.main(Main.java:8)
2020-09-14T17:04:33.0222379Z 
```

noursaidi:
@jhughesbiot what I'm concerned about is differentiating between a "good fail" and a "bad fail". Take the case of a device which only supports TLSv1.3, which should be a good thing. When tested however, the TLSv1.2 test would fail, giving several reasons which would seem to suggest that TLSv1.2 is actually supported, but is an insecure implementation of it.  

This instance of "good fail" and "bad fail" is also something to consider in the future, because it prevents qualifying a device based solely on the pass/fail result and instead needs the failure reason to also be taken into account. I think there needs to be a "does this support the minimum requirements", and then the information specific to the protocol versions are additional information only tests.

With regards to the process, I think it's because it default fails to those conditions, the test report has no information in the case a protocol isn't supported

```Gathering TLS 1 Server Information....
TLS 1 Server Information Complete.

Gathering TLS 1.2 Server Information....
TLS 1.2 Server Information Complete.

Gathering TLS 1.3 Server Information....
Checking Public Key Length...
Validating Public Key
 Sun EC public key, 256 bits
........... (report continues) .......
```

As for the CA resolve errors, it helps that it's in the report now, but I don't think there should be an expectation to read the report when the failure reason give "certificate not signed" is very clear, but unknowingly incorrect? It's an additional concern because at the moment each device needs to have the CA Certificate copied into the device or device type directory, so it's something which is likely to be mis-configured. I reckon because in practice it's the one certificate for all devices, we should see if there's a way to have it "global" so it can be referenced to once for all tests, but perhaps as a future enhancement

jhughesbiot:
@noursaidi 
I'll try to itemize the concerns so we can resolve as needed.

1. Wrongfully reporting support but partial failure:
   This isn't actually the case as it will only partial fail if the test can make a TLS connection at whatever specified version.  If the specified TLS version fails to connect, then you will see results similar to the first tls modules test results indicating a skip.  Only if the connction is succesful does the code continue to validate the elements of that connection.

```
Testing tls alt
RESULT skip security.tlsv1.server IOException unable to connect to server.
RESULT skip security.tlsv1_2.client No client initiated TLS communication detected
RESULT skip security.tlsv1_2.server IOException unable to connect to server.
RESULT skip security.tlsv1_3.client No client initiated TLS communication detected
RESULT skip security.tlsv1_3.server IOException unable to connect to server.
```

2. Minimum requirements pass/fail vs info:
Are you proposing changing the specific version tests to info and aggregating to a singular result for client and server pass/fail results based on a minimum version requirement?  i.e. if TLS 1.2 is the minimum, we pass the TLS serve/client tests only if TLS 1.2 or 1.3 connection can be verified but still report all versions as info?

3. CA certificate issues:
I agree there's room for user error here on the CA installation but there has to be some mechanism to make this happen.  There is the CA_Faux.pem located in the certs directory that could be used as a global file that could be used as a fallback similar to how it is right now in the module_tests.  This has potential for user error though because the tester has to still load a signed cert into the device before testing which I think is going to have similar issues as loading the CA into the device/aux folder.  Not sure the best solution on this one yet but the messaging for this specific CA signature failure could also be updated to be less generic if that was a preferred stop gap to future enhancements.

noursaidi:
@jhughesbiot thanks for itemizing, much clearer this way. 

1. The IOException error only seems to occur when there is no server. I tested the configuration of TLS1.3 only (for example using "Modern" TLS configurations from [Mozilla's SSL configurator here](https://ssl-config.mozilla.org/#server=apache&version=2.4.41&config=modern&openssl=1.1.1d&guideline=5.6)) and the summary table is as follows (see previous comment for report output)
```
|fail|security.tlsv1.server|Other|Other|Certificate public key length too short. Certificate could not be validated. Certificate has not been signed by a CA.|
|skip|security.tlsv1_2.client|Other|Other|No client initiated TLS communication detected|
|fail|security.tlsv1_2.server|Other|Other|Certificate public key length too short. Certificate could not be validated. Certificate has not been signed by a CA.|
|skip|security.tlsv1_3.client|Other|Other|No client initiated TLS communication detected|
|pass|security.tlsv1_3.server|Other|Other|Certificate public key length is >= 224. Certificate active for current date. Certificate has been signed by a CA. Cipher check not required.|
```  

2. That's what I was expecting from the original document, but I have seen the conversations requesting per protocol so I'm wondering if that was something that was missed and should have been there.  I think it can be resolved if the test definition changed (or alternatively interpreted?) from than prescriptive "has a server on port 443 and supports TLSv1.2" it seems to be now, into "supports TLSv1.2 as a server", for example. Then the instance where a protocol is not supported, it could just skip? This would address this and the previous point. Not sure what your thoughts are? @pisuke - thoughts too?

3. Noted, fair point. The overall result (fail) is still correct, so I think it's fine for now and something to decide in the future as mentioned and determine the best solution then

jhughesbiot:
1. That is correct. If TLS 1.2 is supported, then that means a connection will be established.  If you don't have a server that supports TLS 1.2, it will not make the connection and skip.  If the connection is established, then that's when you see itemized failures because that is helpful in indicating what about 1.2 implementation is the failure point and that happens for whatever version is being tested.  Based on a quick review of that config, it looks like it has TLS 1.2 enabled so that's why you're seeing it itemize all the failures.  I believe this is the right behavior as it shows TLS 1.2 is enabled and supported by the device/server but is not implemented to the degree in which it is required to pass based on the itemized failures.  

2. Changing the itemized tests to info and then aggregating to results based on minimum version criteria think is the best way to go.  I'll hold off on any more updates on this until further feedback on a decision we all agree on though.

3. I think we are agreeing here to leave as is until a later date for enhancement.

noursaidi:
I think it can connect because there's a server listening, but if it attempts a TLS1.2 connection then it would fail because that protocol is not supported by the server

The below output is from openssl client which demonstrates this. The _alert protocol version_ means that "The protocol version the client has attempted to negotiate is recognized, but not supported. (For example, old protocol versions might be avoided for security reasons). This message is always fatal. [(source)](https://www.openssl.org/docs/man1.1.1/man3/SSL_alert_type_string.html)" That's why I think the error message and/or result should reflect that the protocol is not supported by the server.  

```
nour@ubun2:~/tls/daq$ openssl s_client --connect 10.42.0.233:443 -tls1_2
CONNECTED(00000005)
140061387186624:error:1409442E:SSL routines:ssl3_read_bytes:tlsv1 alert protocol version:../ssl/record/rec_layer_s3.c:1528:SSL alert number 70
---
no peer certificate available
---
No client certificate CA names sent
---
SSL handshake has read 7 bytes and written 214 bytes
Verification: OK
---
New, (NONE), Cipher is (NONE)
Secure Renegotiation IS NOT supported
Compression: NONE
Expansion: NONE
No ALPN negotiated
SSL-Session:
    Protocol  : TLSv1.2
    Cipher    : 0000
    Session-ID:
    Session-ID-ctx:
    Master-Key:
    PSK identity: None
    PSK identity hint: None
    SRP username: None
    Start Time: 1600123034
    Timeout   : 7200 (sec)
    Verify return code: 0 (ok)
    Extended master secret: no
---
```

jhughesbiot:
@noursaidi 
I'm not sure about that. TLS 1.0 on real devices can't connect to TLS 1.2 only supported devices and is skipped in validation because the java connection forces the TLS standard so it can only connect to valid servers supporting that.  

In that messaging, the alert message indicates a failure on tlsv1, not tlsv1_2.  The remaining information seems to indicate a successful connection on tls1_2 and made a succesful handshake so I don't know if that's the right place to look.

Can you attach the activate.log during your test with the tls1_2 issue?  It may be logging more specific issues around the failure and the result messaging is using default fail values since that is the default state of all those result variables.  This may just be a  better handling around that scenario rather than the connection validation portion.  i.e., if the connection fails to complete a handshake, report it as such instead of itemizing the default failure messages for all the elements of the test.

noursaidi:
activate.log attached as requested. Looking through it, for the unsupported protocols (1.0/1.2), where it says "gathering information" it immediately jumps to "server information complete", and there is also the protocol version alert for these protocols. I think your final sentence with regards to the better handling is all it needs, and it looks like it already detects it so just needs to report it

[activate.log](https://github.com/faucetsdn/daq/files/5221554/activate.log)

jhughesbiot:
@grafnu 
I managed to get the module tests to work but am still experiencing an odd behavior between local and CI builds.  This was actually a failure i've been seeing for a while and don't see a way to account for it.  If you look at this run in the test_aux https://github.com/faucetsdn/daq/pull/626/checks?check_run_id=1114795692, you'll see the results mismatching the section shown below, which is what is generated when I run locally.  Local run and CI run generating different test.out files makes it difficult to get a good passing success and requires a lot of retries. Hope you have some ideas on this.

```
dp_sec_port_1_acl.yaml
dp_sec_port_2_acl.yaml
dp_sec_port_3_acl.yaml
dp_sec_port_4_acl.yaml
dp_sec_port_5_acl.yaml
dp_sec_port_6_acl.yaml
```

grafnu:
Just to clarify, when you run locally you see that it does generate the
"dp_sec_port_6_acl.yaml" file, but the CI runs do not see this file?

Can you try deleting the inst/port_acls directory and see if it changes
your local results?  Or, look at the timestamps of the files in
inst/port_acls, and see if the dp_sec_port_6_acl.yaml file is
significantly older than the others.  My suspicion is that they aren't
being cleared out for whatever reason, and so you're getting local errors
b/c of cruft from a previous run of something.  If this is the case, then
the fix is simple (add a rm command to the testing script), but I want to
verify first.

I think the root cause is the "bin/mudacl" script that does clear this out
-- at one point it was in test_aux.sh but got moved to test_topo.sh...
Doesn't matter on a clean build (which is why CI doesn't catch it), but
might be causing you grief.

jhughesbiot:
@grafnu 
Yes, sec_port_4 through 6 show up locally but not in the CI.  I checked in the inst directory and see all those configs so I cleared out the entire inst directory and re-ran again and now it all matches.  I guess there's some extra cleanup that needs to happen locally or added into the testing script as you mentioned.

jhughesbiot:
@noursaidi 
Quick recap on the outstanding items:

1. Add skip scenario for failed handshake to account for unsupported TLS versions
2. Change current tests to info and add two more results with pass/skip/fail based on highest version for both client and server results.  This would change the current results to something like:
@pisuke Please add any feedback on this you think is necessary.
```
Testing tls alt tls
RESULT info security.tlsv1.server Certificate public key length is >= 224. Certificate active for current date. Certificate has been signed by a CA. Cipher Valid.
RESULT info security.tlsv1_2.client No client initiated TLS communication detected
RESULT info security.tlsv1_2.server Certificate public key length is >= 224. Certificate active for current date. Certificate has been signed by a CA. Cipher Valid.
RESULT info security.tlsv1_3.client No client initiated TLS communication detected
RESULT info security.tlsv1_3.server Certificate public key length is >= 224. Certificate active for current date. Certificate has been signed by 
RESULT pass security.tls.server TLS server supported at versions: 1.0, 1.2 and 1.3.
RESULT skip security.tls.client No client initiated TLS communication detected
```

3. No action, future enhancement.

noursaidi:
@pisuke @spectre-storm do you agree with the above?

grafnu:
https://github.com/faucetsdn/daq/pull/640

> @grafnu
> Yes, sec_port_4 through 6 show up locally but not in the CI. I checked in the inst directory and see all those configs so I cleared out the entire inst directory and re-ran again and now it all matches. I guess there's some extra cleanup that needs to happen locally or added into the testing script as you mentioned.

jhughesbiot:
@noursaidi 
Update is implemented to account for skip results for bad handshakes

@pisuke @spectre-storm 
Looking for feedback to get a consensus on current reporting and updating in future or handling it now with an agreed upon approach.

@grafnu 
Last build failed in aux but looks like some other issue around DHCP reporting.  I can't force a rerun of this test to confirm the issue so hopefully you or somebody else with the correct permissions can trigger a rerun.

grafnu:
I'm away from my computer so can't re-trigger, but you can always push an
empty commit...

pisuke:
> @pisuke @spectre-storm
> Looking for feedback to get a consensus on current reporting and updating in future or handling it now with an agreed upon approach.

I think the current reporting meets the minimum requirements to understand the TLS implementation on the device.
Once it passes the CI tests, let's approve and merge and then address other issues after having trialled the test on a few devices and gathered user feedback.

pisuke:
> I'm away from my computer so can't re-trigger, but you can always push an empty commit...

I've just re-triggered.

jhughesbiot:
@pisuke @grafnu 
Re-running of the tests seemed to get them all to pass but now there are merge conflicts that I don't seem to have access to review and fix.  Resolve conflicts button is not accessible to do directly and the conflicts do not show up locally.

grafnu:
It says in there you have to use the command line to resolve (I don't know
why either, but the button is disabled for me too).

noursaidi:
@jhughesbiot you could try merging the latest changes from the master branch of this repo into your local branch. It should show a merge conflict but you'll be able to address it that way.

jhughesbiot:
@noursaidi 
Good call on the master.  Something changed with that merge causing an issue with the modules test but i think i found it just waiting for a clean build before we can close this out.

jhughesbiot:
@noursaidi @grafnu @pisuke 
builds passing now so we should be good on this

grafnu:
All good from my side (already approved)... not sure if there were other outstanding concerns.

> @noursaidi @grafnu @pisuke
> builds passing now so we should be good on this

pisuke:
No other concerns for this iteration, I'll press the merge button. Thanks!

PR #634, opened 2020-09-10
Demo merging ipaddr and dhcp

henry54809:
b/168020741
The current ipaddr module, if enabled, will automatically kick off the dhcp container directly in the module when all tests have finished.
This also comes with a dummy dhcp module to demo the functionality. It is meant to be replaced by content in the 2 DHCP PRs.


subset/ipaddr/Dockerfile.test_ipaddr
```diff
@@ -0,0 +1,8 @@
+FROM daqf/aardvark:latest
+
+RUN $AG update && $AG install python python-setuptools python-pip netcat
+
+COPY subset/ipaddr/dhcp_tests.py .
+COPY subset/ipaddr/test_dhcp .
+
+CMD ["./test_dhcp"]
```
grafnu, line 8: should generally have newlines at end of file

grafnu (commented):
Mostly fine -- just a nit and stickly/test stuff...

henry54809:
PTAL. Tests passed.

subset/ipaddr/Dockerfile.test_ipaddr
grafnu: this one's fixed, but there are others...

henry54809: Actually, I looked through all Dockerfiles, conf files and python files and they don't(except for 1 conf file) have new lines at the end of the file. I want to keep them consistent; should I remove this new line, or add new lines to all files?

grafnu:
They should all have newlines. I wouldn't necessarily take that on as part
of this PR, as such -- since it will conflate two things. If you want to
modify everything to be consistent that should really be done separately...

PR #647, opened 2020-09-18
fix no_test option

henry54809:


grafnu (approved):
Well that was simple!

PR #648, opened 2020-09-19
Basic UDMI provisioner test

grafnu:


daq/report.py
```diff
@@ -172,13 +172,16 @@ def _process_results(self):
...
             path = result_dict[ResultType.REPORT_PATH]
+            results = []
             with open(path) as stream:
                 for line in stream:
                     match = re.search(self._RESULT_REGEX, line)
                     if match:
-                        result, test_name, extra = match.group(1), match.group(2), match.group(3)
-                        self._accumulate_result(test_name, result, extra, module_name=module_name)
-                        module_result["tests"][test_name] = self._results[test_name]
+                        name, result, description = (match.group(2), match.group(1), match.group(3))
```
henry54809, line 180: I don't understand why this change is made? The 2 seems to be equivalent, except this is looped one more time?

grafnu: Not entirely sure myself. I think I did that early on when I was debugging something for some reason. But, two loops are not better than one, so reverted.

grafnu:
Updated, PTAL.

henry54809 (approved):

PR #651, opened 2020-09-22
Tls timeout fix

jhughesbiot:


subset/security/tlstest/src/main/java/Server.java
```diff
@@ -508,10 +509,9 @@ private String getSessionCipher(SSLSocket socket) throws IOException {
   private SSLSocket makeSSLSocket(String host, int port, String protocol)
       throws NoSuchAlgorithmException, KeyManagementException, IOException {
     SSLSocketFactory factory = makeSSLFactory(trustAllManager(), protocol);
-
-    SSLSocket socket = (SSLSocket) factory.createSocket(host, port);
+    SSLSocket socket = (SSLSocket)factory.createSocket();
+    socket.connect(new InetSocketAddress(host, port), 10000);
```
grafnu, line 513: NIT: Ideally extract to named constant so there's some contextual meaning, e.g.:

CONNECT_TIMEOUT_MS

grafnu (approved):

subset/security/tlstest/src/main/java/Server.java
jhughesbiot: Easy enough.

PR #652, opened 2020-09-22
ipaddr DHCP timeout behavior change

henry54809:


daq/test_modules/ipaddr_module.py
```diff
@@ -130,6 +130,10 @@ def ip_listener(self, target_ip):
...
         if self._timeout and datetime.now() >= self._timeout:
-            self._logger.error('DHCP times out after %ds lease time.' % self._lease_time_seconds)
-            self.terminate()
-            self.callback(exception=self._TIMEOUT_EXCEPTION)
+            if self.docker_host.start_time:
+                self.terminate()
+                self.callback(exception=self._TIMEOUT_EXCEPTION)
+            else:
+                self._logger.error('DHCP times out after %ds lease time' % self._lease_time_seconds)
+                self.tests = self.tests[-1:]
```
grafnu, line 138: can you extract the [-1:] to a function and then also use it where the "self.tests.pop(0)" happens?   Seems like both cases are the same -- "pop the top test off the stack" -- just in one case the return value would be ignored.

grafnu (approved):

daq/test_modules/ipaddr_module.py
henry54809: The 2 are quite different, so I don't know if a function is beneficial. self.tests[-1:] is for removing all tests before the analysis tests but the return value is still an array. self.tests.pop(0) is to take the head, so the return value is one element of the array.

grafnu:
Oh right -- my bad.  I was getting confused by [-1:] (thought it was [1:]

PR #656, opened 2020-09-24
Add gRPC client to send device testing result to server

didovesei:


grafnu (commented):
We need to find a way to make the codecov warnings go away b/c the file is unreviewable in its current state... Not sure what's involved, but we may have to disable codecov if it's too problematic (at least until the PR is reviewed).

henry54809:
It's probably because all the tests failed. Once that's resolved, the warnings should go away. 
> We need to find a way to make the codecov warnings go away b/c the file is unreviewable in its current state... Not sure what's involved, but we may have to disable codecov if it's too problematic (at least until the PR is reviewed).

henry54809 (approved):

bin/setup_dev
```diff
@@ -168,6 +169,11 @@ else
     (cd forch; git fetch; git checkout $FORCHV)
 fi
 
+SITE_PACKAGES_PATH=$(python3 -c 'import sysconfig; print(sysconfig.get_paths()["purelib"])')
```
grafnu, line 172: where is this?  Is this in /usr/lib or something like that?  If so, we really shoudn't be futzing with it...

daq/device_testing_client.py
```diff
@@ -0,0 +1,23 @@
...
+import grpc
+
+from forch.proto.grpc.device_testing_pb2_grpc import DeviceTestingStub
+from forch.proto.device_testing_state_pb2 import DeviceTestingState
+
+_SERVER_ADDRESS_DEFAULT = '127.0.0.1'
+_SERVER_PORT_DEFAULT = 50051
+
+
+class DeviceTestingClient:
```
grafnu, line 12: don't use Testing

proto/system_config.proto
```diff
@@ -126,6 +126,9 @@ message DaqConfig {
 
   // Drop into console mode after test
   bool use_console = 52;
+
+  // Port of the server to receive device testing result
+  bool device_testing_server_port = 53;
```
grafnu, line 131: again don't use testing -- device_result_port would be fine (is server redundant, there's never a client port...?)  What do we have elsewhere where a port is specified?

bin/setup_dev
didovesei: Reverted this file. Seems that the change is not useful..

daq/device_testing_client.py
didovesei: Changed to DeviceQualificationClient

proto/system_config.proto
didovesei: This is the only place in the config where there is a port. Changed to device_result_port.

testing/unit/test_runner.py
```diff
@@ -79,42 +77,45 @@ def test_reap_stale_ports(self):
         self.runner.target_set_error.assert_called()
 
 
-class TestRunnerSendingTestingResult(TestRunnerBase):
-    """Test case to test the runner sending device testing result to the device testing server"""
+class TestRunnerSendingQualificationResult(TestRunnerBase):
```
grafnu, line 80: This class name is too specific for a testing class.  This would be an individual test ("test_runner_sending_qualification_result") -- for the class name it should be a general bucket that encompasses a few things.  So, what's the next level up from this specific use case? What's common to all the helper methods and data that's in here?

daq/device_qualification_client.py
```diff
@@ -9,7 +9,7 @@
 _SERVER_PORT_DEFAULT = 50051
 
 
-class DeviceTestingClient:
+class DeviceQualificationClient:
```
grafnu, line 12: Qualification is not right either. That's a description of the USE of it, not what it is.  Also, this entire repo is "qualification" so this name could apply to any client that's in the repo.   What is it connecting to on the other end?  B/c it's a client of that and that's what should be in the name.

didovesei: It is connecting to `DeviceTestingServer` in forch... But I think that should be renamed as well... What about `DevicePortBehaviorClient` or `DeviceResultClient`?

testing/unit/test_runner.py
didovesei: Yes, makes sense. The same test case can be reused when target devices complete the tests with different results (passed, error, etc.). What about TestTargetDeviceComplete?

grafnu:
Can you maybe draw a dependency diagram of the "PortBehavior.Behavior''
enum in forch -- just a class/proto hierarchy that shows all the different
"sources" for the port behavior. I think that will make it more clear
overall and give us some consistency as to the naming. I was taking a quick
look there and I"m seeing a lot of redundancy so I want to give it a pass
to clean it up while we're thinking about it!

testing/unit/test_runner.py
grafnu: Oh, I see my confusion -- this class shouldn't be in this file at all. This class is specific to a kind of test so should be in an appropriate module file. This module (test_runner) is really only for the basics of running all kinds of tests, so shouldn't have any dependency on a specific (e.g. DeviceTestingServer).

grafnu (commented):
biggest comment would be that I think a proto diagram for the forch server would be a good next step... I think we need a big picture of all the things that lead to a PortBehavior specification before we can really understand the right name.

didovesei:
I have udpated the class names to match the gRPC protos in Forch. PTAL..

proto/system_config.proto
```diff
@@ -126,6 +126,9 @@ message DaqConfig {
 
   // Drop into console mode after test
   bool use_console = 52;
+
+  // Port of the server to receive device result
+  bool device_result_port= 53;
```
grafnu, line 131: - bool?  Shouldn't this be an int?
- Space before =
- Encapsulate in submessage (e.g. something new similar to RunTrigger). I.e., what if we (in the future) also wanted to specify an ip address?

daq/runner.py
```diff
@@ -166,6 +169,9 @@ def __init__(self, config):
         self._system_active = False
         logging_client = self.gcp.get_logging_client()
         self.daq_run_id = self._init_daq_run_id()
+        self._device_result_client = None
```
grafnu, line 172: little bit clumsy doing it this way, best to assign the variable (_device_result_client) only in one place. E.g., 
self._device_result_client = self._init_device_result_client(self.config.get('device_result_port'))

daq/runner.py
```diff
@@ -195,6 +201,10 @@ def _init_daq_run_id(self):
             output_stream.write(daq_run_id + '\n')
         return daq_run_id
 
+    def _init_device_result_client(self):
+        self._device_result_client = DeviceReportClient(
+            server_port=self.config['device_result_port'])
```
grafnu, line 206: best to reference device_result_port only in one spot (similar to other comment)

proto/system_config.proto
didovesei: Encapulsated in DeviceReporting.

daq/runner.py
didovesei: updated

daq/runner.py
didovesei: Removed the one in __init__().

grafnu (approved):

PR #658, opened 2020-09-28
IP Change (11)

frgitdaq:


docker/include/network/scripts/new_ip
```diff
@@ -36,4 +36,7 @@ else
     new="dhcp-host=$mac_addr,$new_ip"
     flock /etc/dnsmasq.conf echo $new >> /etc/dnsmasq.conf
 fi
+
+source change_dhcp_range $(echo $new_ip | cut -d '.' -f 1,2,3).0 $(echo $new_ip | cut -d '.' -f 1,2,3).255 24
```
frgitdaq, line 40: @henry54809 Looking back at the test_dhcp runs on master, it looks as if the new_ip script does not work if used alongside the multi-subnet test. This is because the device is sending a DHCP request which is accepted by dnsmasq (even though a static host has been set) because the static host is outside of any range registered in the dnsmasq config. I tried adding a static range like this: `dhcp-range=10.0.0.0,static` but this was not successful. The current method is successful however by changing the whole range.

henry54809: Can you please revert the changes here. I will fix this in a different PR.

henry54809 (approved):

subset/ipaddr/dhcp_tests.py
```diff
@@ -79,6 +82,32 @@ def _test_dhcp_short():
         fd.close()
         return 'fail', 'No DHCP request received.'
 
+    def _test_ip_change():
```
henry54809, line 85: There should be an integration test to test the output of this script.

frgitdaq: > There should be an integration test to test the output of this script.

@henry54809 Is that not what device 05 in test_dhcp is for? Otherwise, are you suggesting a test_aux case?

henry54809: That integration only runs this function but it doesn't verify the output of the function. There should be something to check for generated reports after the run finishes in the existing dhcp integration test.

frgitdaq: @henry54809 Does the latest commit seem like what you had expected?

testing/test_dhcp.sh
```diff
@@ -92,17 +92,26 @@ cmd/run -b -s settle_sec=0 dhcp_lease_time=120s
...
     num_ips=$(cat $ip_file | cut -d ' ' -f 1 | sort | uniq | wc -l)
+    ip_change=$(cat $report_file | fgrep 'pass connection.dhcp.ip_change' | wc -l)
+    dhcp_timeouts=$(cat inst/cmdrun.log | fgrep 'DHCP times out after 120s lease time' | wc -l)
     echo Found $ip_triggers ip triggers and $long_triggers long ip responses.
-    if [ $iface == 5 ]; then
+
+    if [ $iface == 6 ]; then
+      device_dhcp_timeouts=$(cat inst/cmdrun.log | fgrep 'DHCP times out after 120s lease time' | fgrep "ipaddr_ipaddr0$iface" | wc -l)
+      echo "Device $iface DHCP timeouts: $device_dhcp_timeouts" | tee -a $TEST_RESULTS
+      echo "Device $iface ip change: $((ip_change == false))" | tee -a $TEST_RESULTS
```
henry54809, line 110: That should be $ip_change and the output should be 0 to indicate there isn't an IP change to avoid confusions.

testing/test_dhcp.sh
```diff
@@ -114,10 +123,8 @@ for iface in $(seq 1 5); do
     else
       echo "Device $iface ip triggers: $((ip_triggers > 0)) $((long_triggers > 0))" | tee -a $TEST_RESULTS
     fi
+
 done
 
-dhcp_timeouts=$(cat inst/cmdrun.log | fgrep 'DHCP times out after 120s lease time' | wc -l)
-device_6_dhcp_timeouts=$(cat inst/cmdrun.log | fgrep 'DHCP times out after 120s lease time' | fgrep 'ipaddr_ipaddr06' | wc -l)
 echo "DHCP timeouts: $dhcp_timeouts" | tee -a $TEST_RESULTS
```
henry54809, line 129: This can be removed.

frgitdaq: @henry54809 were you referring to the total DHCP timeouts?

henry54809: That check was supposed to be for device 6, but you have included device 6 results in the for loop before this already.

PR #662, opened 2020-09-30
gateway refactoring

henry54809:


daq/runner.py
```diff
@@ -13,17 +13,20 @@
 
 import configurator
 import faucet_event_client
-import gateway as gateway_manager
+import container_gateway
+
+# To be uncommented as part of the next PR
```
grafnu, line 18: nit: standard way of saying this is with a TODO

# TODO: To be uncommented as part of external gateway PR

TODO has a significance, and useful to indicate what it is (if it's worth having the comment might as well give it meaning)

grafnu (approved):
minor nit and a stickler, I think... but all good from my end.

PR #664, opened 2020-10-03
External Gateway

henry54809:


cmd/auto
```diff
@@ -41,7 +41,7 @@ egrep $regex $CFILE | while read iopt; do
     key=${BASH_REMATCH[2]}
     val=${BASH_REMATCH[3]}
     if [ "$key" = opts ]; then
-        start_faux $iface $(eval echo $val)
+        start_faux $iface $val
```
henry54809, line 44: I'm not sure what the eval was intended for but it's not passing the the "-n" option correctly.

grafnu: Seems like the kind of thing that would benefit from a comment as to why it's there! (Likely my fault).

For now just do whatever you need for the tests to pass and/or it to work, and then if we run into the situation where it's required again. And then make sure to add the comment.

config/system/alt.yaml
```diff
@@ -18,6 +18,10 @@ interfaces:
   faux:
     opts: 
     port: 2
+  faux-3:
```
grafnu, line 21: this shouldn't mix faux and faux-3 -- so should change the first one to faux-1

daq/runner.py
```diff
@@ -578,8 +578,6 @@ def _activate_device_group(self, device):
         LOGGER.info('Gateway for device group %s not found, initializing base %d...',
                     device.group, set_num)
         if device.dhcp_mode == DHCPMode.EXTERNAL:
-            # TODO to be removed.
```
grafnu, line 581: roughly indicate why

henry54809:
PTAL.

daq/runner.py
grafnu: The TODO should indicate why -- also the comment you added doesn't explain "why" -- it's saying what that line is doing.  Why should this be removed eventually -- i.e. under what condition should it be removed?

henry54809:
This TODO line was already removed in this PR so I wasn’t really sure what
comment you were referring to to be put in there.

grafnu (approved):
Oh, apparently I was confused (thought you were adding the TODO). I'm color-blind so the color-coding of removed-vs-added isn't very significant and I just didn't think about the -/+ part.

PR #665, opened 2020-10-05
DHCP Change (20)

frgitdaq:


frgitdaq:
Putting into draft until PR #658 is merged due to some overlap.

daq/test_modules/ipaddr_module.py
```diff
@@ -102,6 +103,15 @@ def _ip_change_test(self):
         self.host.gateway.request_new_ip(self.host.target_mac)
         self._ip_callback = self._next_test
 
+    def _dhcp_change_test(self):
+        self._set_timeout()
+        if not self.host.connect_port(False):
+            self._logger.error('disconnect port not enabled')
```
henry54809, line 109: I thought for this test you shouldn't disconnect port, so that it can test if the device can accept a new IP from the dhcp server.

henry54809: Nevermind my last comment. I thought it was ip_change test.

subset/ipaddr/dhcp_tests.py
```diff
@@ -91,6 +93,18 @@ def _test_private_address():
             return 'pass', 'All private address ranges are supported.'
         return 'fail', 'Not all private address ranges are supported.'
 
+    def _test_dhcp_change():
```
henry54809, line 96: There should be an integration test to test the output of this script.

noursaidi:
@frgitdaq @pisuke @spectre-storm Does this need to check that the device is pingable after the IP change like test no. 11? The test document is not clear but I think it should perform this check

frgitdaq:
> @frgitdaq @pisuke @spectre-storm Does this need to check that the device is pingable after the IP change like test no. 11? The test document is not clear but I think it should perform this check

@noursaidi It certainly wouldn't harm adding the additional check for the ping response. I have added this.

henry54809:
I understand @pisuke will rename this test to dhcp.disconnect_ip_change? That's a better name to describe this test than dhcp_change.

henry54809 (approved):

pisuke:
Yes, I'll rename the test according to the latest test plan.

PR #671, opened 2020-10-08
set gateway set vlan

henry54809:


config/system/alt.yaml
```diff
@@ -22,6 +22,10 @@ interfaces:
     opts: -n 
     port: 3
 
+default_timeout_sec: 120
+
 # use vlan trigger
 run_trigger_type: VLAN
-default_timeout_sec: 120
+test_vlan_range:
```
grafnu, line 29: We've been avoiding the prefix "test" for this on principle since it's very confusing since it conflicts with the 200 other uses of the word "test" we have floating around -- so we should find something else to demark this.  Also "vlan" is redundant and unnecessarily specific, since we could, in theory, have other things that would trigger this.

I'm realizing now that the "run_trigger_type" param is also redundant, since it can be inferred from the existence of a vlan range.

How about we go with a slightly generalized form that looks like:

run_trigger:
  vlan_start: 2000
  vlan_end: 2100

and remove the "run_trigger_type" param and have that be inferred?

I'm somewhat future-proofing and generalizing to the world where there might be other and/or multiple triggers. Not sure what that might be, but conceptually there's nothing that would prevent different triggers from existing (e.g. based on MAC address, or some other parameter)... and keeping this more general would allow for a convenient bucket for all that.

daq/topology.py
```diff
@@ -1,5 +1,4 @@
 """Faucet-specific topology module"""
-
```
grafnu, line 2: prevailing code convention is to have the blank line after the initial comment... was something complaining about this?

proto/system_config.proto
```diff
@@ -126,6 +126,9 @@ message DaqConfig {
 
   // Drop into console mode after test
   bool use_console = 52;
+
+  // Test vlan range
+  TestVLANRange test_vlan_range = 53;
```
grafnu, line 131: As mentioned previously, I think this should be renamed to something more generic.

(Also, as an aside this would be TestVlanRange since in this context it's equivalency between snake_case and CaMeL case that matters.)

henry54809:
PTAL

grafnu (approved):

PR #699, opened 2020-10-29
Cleanup vlan handling in prep for fot integration

grafnu:


daq/topology.py
```diff
@@ -36,7 +36,8 @@ class FaucetTopology:
     _MIRROR_IFACE_FORMAT = "mirror-%d"
     _MIRROR_PORT_BASE = 1000
     _SWITCH_LOCAL_PORT = _MIRROR_PORT_BASE
-    _VLAN_BASE = 1000
+    _LOCAL_VLAN = 1000
+    _DUMP_VLAN = 999
```
henry54809, line 40: What is this dump vlan for?

grafnu:
Maybe NULL or DROP vlan would be better (but didn't want to get confused to
None)... This is basically a "nothing" vlan that should drop all traffic.
I think (but maybe I should check) if you don't set some kind of vlan on
the port then Faucet will barf.   It's basically a "disconnected" setting.
 Previously it used to be 1000 (_VLAN_BASE), but with the potential
introduction of the vlan range triggers I wanted to make it more explicit.
I'm trying to do this in 2-phases, and this phase is more of a refactoring
with not too many changes, and then next I need to do some things like make
this configurable, etc...

PR #701, opened 2020-10-30
Dynamically generate dhcp reflection for vlan triggers

grafnu:


docker/include/network/scripts/start_networking
```diff
@@ -40,8 +40,13 @@ if ! ip addr show dev $LOCAL_IF | fgrep -q 'inet '; then
   ip addr add 10.20.$subnet.1/16 dev $LOCAL_IF
 fi
 
-ip addr add 10.20.255.254/16 dev $LOCAL_IF #For static ip devices' default gateway IP
-echo dhcp-host=*,ignore >> /etc/dnsmasq.conf
+# For static ip devices' default gateway IP
+ip addr add 10.20.255.254/16 dev $LOCAL_IF
+
+# Block execution until first host attaches unless being used as a faux device.
```
henry54809, line 46: I think this may be breaking the alt-swtich integration test. The integration tests waits for a log message before setting the dhcp range.

grafnu: I don't think it was this that was breaking the tests, there were other more fundamental problems.  But, either way this is no longer necessary given the new mechanism.  Still running a few final tests but I think I got everything now...

daq/topology.py
```diff
@@ -42,6 +42,7 @@ class FaucetTopology:
     PRI_TRUNK_NAME = 'trunk_pri'
     _NO_VLAN = "0x0000/0x1000"
     _EXT_STACK = 'EXT_STACK'
+    _IN_PORT = 0xfffffff8
```
henry54809, line 45: Maybe we should use the the full name OFPP_IN_PORT here. in_port is a bit confusing on its own.

docker/include/network/scripts/start_networking
grafnu: Should be fixed now.

daq/topology.py
grafnu: Agreed. Done.

grafnu:
Took way more than expected, but I fixed all the tests. PTAL.

henry54809 (approved):

PR #710, opened 2020-11-13
Adding a IP that falls in device ip's subnet to test hosts

henry54809:


daq/test_modules/native_host.py
```diff
@@ -42,6 +42,11 @@ def activate(self, log_name='activate.log'):
         assert not self.active_pipe, '%s already activated' % self.name
 
         env = dict(self.env_vars)
+        if self.intf() and "TARGET_IP" in env and not env["TARGET_IP"].startswith('10.20'):
+            parts = env["TARGET_IP"].split('.')
+            parts[-1] = str((int(parts[-1]) + 1) % 256)
```
didovesei, line 47: I think it's better to change the parts[-2] instead of incrementing parts[-1]... During the test I can see that faux-1 is assigned by the DHCP server with 192.168.1.207/16 and faux-2 with 192.168.1.208/16. Setting the IP of the testing host for faux-1 to 192.168.1.208 would cause an IP conflict... Since the subnet is /16, altering the last second byte should not still work..

didovesei (approved):

PR #714, opened 2020-11-19
Add external DHCP subnet config

henry54809:


config/system/alt.yaml
```diff
@@ -30,3 +30,12 @@ run_trigger:
...
   vlan_end: 1009
   egress_vlan: 121
+
+mininet_subnet:
+  base_ip: 192.168.1.0
+  prefix_length: 24
+
+# Define external DHCP subnets to be used by test hosts
+external_dhcp_subnets:
+  - base_ip: 10.20.99.0
```
grafnu, line 40: I would go with the CIDR nomenclature... it's much more common and concise:

10.20.99.0/24

config/system/alt.yaml
```diff
@@ -30,3 +30,12 @@ run_trigger:
   vlan_start: 1001
   vlan_end: 1009
   egress_vlan: 121
+
+mininet_subnet:
+  base_ip: 192.168.1.0
+  prefix_length: 24
+
+# Define external DHCP subnets to be used by test hosts
+external_dhcp_subnets:
```
grafnu, line 39: maybe just "external_subnets" -- I'm not sure "DHCP" matters... e.g. it applies to static IPs as well.

config/system/alt.yaml
```diff
@@ -30,3 +30,12 @@ run_trigger:
   vlan_start: 1001
   vlan_end: 1009
   egress_vlan: 121
+
+mininet_subnet:
```
grafnu, line 34: What's this "for"?  "mininet" isn't meaningful unless you know the internals (breaks abstraction).  I'm honestly trying to remember myself what this is used for... Maybe `internal_subnet` -- just to differentiate from `external_subnets`?

daq/test_modules/external_module.py
```diff
@@ -46,6 +48,10 @@ def start(self, port, params, callback, finish_hook):
 
         try:
             pipe = host.activate(log_name=None)
+            # for devcies with ips that are not in the same subnet as test hosts' ips.
```
grafnu, line 51: Nit: Start with Capital letter.

henry54809:
PTAL

config/system/default.yaml
```diff
@@ -46,7 +46,5 @@ usi_setup:
   url: localhost:5000
   rpc_timeout_sec: 20
 
-# default mininet subnet
-mininet_subnet:
-  base_ip: 10.20.0.0
-  prefix_length: 16
+# default test host subnet
```
grafnu, line 49: is this default or internal?  They have slightly different meanings -- just wondering about congruence.  I'm fine either way (don't actually know what it's used for), but would at least want the comment to match the field :-)

daq/network.py
```diff
@@ -55,10 +57,8 @@ def __init__(self, config):
         self.sec_dpid = None
         self.sec_port = None
         self._settle_sec = int(config['settle_sec'])
-        subnet = config.get('mininet_subnet')
-        subnet_str = "%s/%s" % (subnet['base_ip'], subnet['prefix_length']
-                                ) if subnet else DEFAULT_MININET_SUBNET
-        self._mininet_subnet = ip_network(subnet_str)
+        subnet = config.get('mininet_subnet', self.DEFAULT_MININET_SUBNET)
```
grafnu, line 60: mininet_subnet?

proto/system_config.proto
```diff
@@ -130,11 +130,11 @@ message DaqConfig {
   // Reporting device result
   DeviceReporting device_reporting = 53;
 
-  // For specifying ip subnets assigned by external DHCP server
-  repeated IpSubnet external_dhcp_subnets = 54;
+  // For specifying ip subnets assigned by external DHCP server / static ips
+  repeated string external_subnets = 54;
```
grafnu, line 134: in terms of future-proofing, can you think of any potential information (properties) we'd maybe need to keep along with the subnet?  E.g., DNS server or some other property? I can't specifically think of something -- just asking the question to make sure we're considering things.  In general, having this be an object (not necessarily IpSubnet but maybe something like SubnetSpec) makes the system more flexible in the future since we can then add a property to each entry. If it's encoded as a string, then there's no path forward for adding annotations.

In the past, every time I've done something like "a list of strings" then I end up regretting it because I often then want to add some kind of information to it and then you're SOL.

config/system/default.yaml
henry54809: This matches the default specified in code so I put the default in comment here. This is used when instantiating a mininet network. http://mininet.org/api/classmininet_1_1net_1_1Mininet.html#a1ed0f0c8ba06a398e02f3952cc4c8393
I thought of 1 issue where test hosts and devices' IPs may conflict. If mininet assigns test host an IP (it usually just increments from the subnet) and external DHCP server assigns the same IP to a device; this was why I made the mininet's subnet configurable.

grafnu: Re: default vs. internal -- it's very confusing as written so please update to make it more clear.

Don't the test hosts have to be on the same subnet as the device?  If not, then the device will try to send the test host traffic to the gateway mac address which would fail?  I guess I don't understand the scenario... Might need a conversation outside of this PR (not sure it directly impacts this PR either).

henry54809:
PTAL

grafnu (approved):

PR #717, opened 2020-11-21
Update report aligning to the mockup

noursaidi:
Issue #708 - Updates the category table to report to match the agreed report mockup. Example of new table below. Also adds a new warning when a test is gone to alert to possible misconfiguration 

Some items which cropped up and may need addressing in the future is the `'required':'pass'` 

---
**Some tests report as GONE. Please check for possible misconfiguration**
|Category|Total Tests|Result|Required Pass|Required Pass for PoE Devices|Required Pass for BACnet Devices|Recommended Pass|Information|Other|
|---|---|---|---|---|---|---|---|---|
|Connection|9|FAIL|1/4/4|0/0/0|0/0/0|0/0/0|0/0/0|0/0/0|
|Security|8|FAIL|1/0/4|0/0/0|0/0/0|1/1/0|0/0/1|0/0/0|
|Network Time|2|PASS|2/0/0|0/0/0|0/0/0|0/0/0|0/0/0|0/0/0|
|TLS|0|N/A|0/0/0|0/0/0|0/0/0|0/0/0|0/0/0|0/0/0|
|Protocol|2|FAIL|0/0/0|0/0/0|0/1/0|0/0/0|0/0/1|0/0/0|
|PoE|3|FAIL|0/0/0|0/1/1|0/0/0|0/0/0|0/1/0|0/0/0|
|BOS|1|SKIP|0/0/0|0/0/0|0/0/0|0/0/1|0/0/0|0/0/0|
|Other|2|GONE|0/0/0|0/0/0|0/0/0|0/0/0|0/0/0|0/2/0|
|Communication|2|GONE|0/1/0|0/0/0|0/0/0|0/0/0|0/1/0|0/0/0|

Syntax: Pass / Fail / Skip


daq/report.py
```diff
@@ -246,33 +252,102 @@ def _write_test_tables(self):
...
-                    required_result = test_info['required']
                     total += 1
-                    if result_dict["result"] == required_result:
-                        match += 1
+
+                    expected_name = test_info.get('expected', self._DEFAULT_EXPECTED)
+                    expected_index = self._expected_headers.index(expected_name)
+
+                    # Put test results into right location in the result matrix
+                    if result_dict["result"] in ("pass", "info"):
```
pisuke, line 323: @noursaidi thanks for putting this together.
About the `info` result output, I think we agreed to remove it.
OK to do this in a separate PR though.

pisuke (approved):
Only one minor comment about the `info` output, but looks good to be merged.

daq/report.py
noursaidi: @pisuke correct, we have issue #712 to cover the removal of info tests which requires requires more work and modification to various test modules. The reference to info tests would be removed as a part of those changes. Until then though, this ensures they are still continue to be accounted for as they are currently are

PR #732, opened 2021-01-07
Fix ping test for vlan trigger runs

henry54809:


daq/test_modules/external_module.py
```diff
@@ -102,6 +103,7 @@ def opt_param(key):
             ("GATEWAY_IP", params['gateway_ip']),
             ("GATEWAY_MAC", params['gateway_mac']),
             ("LOCAL_IP", opt_param('local_ip')),
+            ("EXTERNAL_DHCP", '1' if external_dhcp else '')
```
grafnu, line 106: Does it make sense to keep this as an enum rather than boolean?  So, "DHCP_MODE" with a value of "external"?  Just trying to think of future-proof and keep parity with the self.device.dhcp_mode variable.

grafnu (approved):

PR #733, opened 2021-01-08
Add module include capability

grafnu:


henry54809:
What's the intention of this additional inheritance model? considering DAQ already uses the base<-site<-device type<- device inheritance? it seems like there can easily be an infinite recursion in the merge_config.

grafnu:
It's basically to address the mess that shows up in this
https://github.com/faucetsdn/daq/pull/727 -- the same information is
encoded in multiple places, which makes it harder to keep everything in
sync. The explicit inheritance model is instantiated by function, while
this is meant to be something extrinsic that can be specified as per
specific project/configuration needs.  Yes, there could be infinite
recursion, but I wouldn't call it "easy" as such -- but it would also be
pretty easy to track down and fix (and/or block if necessary).

noursaidi:
@pisuke has been looking into documenting the configuration and is back tomorrow, but makes sense to me - we'd be able to avoid duplicating test descriptions which was the current problem.

The way I think we'd implement this would be to create a `common_tests.json` file which describes the tests shared by remediation and qualification configurations. Then we create a `qualification_tests.json` and `remediation_tests.json` which includes `common_tests.json`, then describes any unique respective configurations.  Then, the `<site path>/module_config.json` includes the appropriate  file instead of repeating the information. We'd also update the CI test site to include the files so it's all in sync.

The only issue which comes to mind is updates. This becomes helpful because keeping the module_config files up to date as new tests have been added and tests renamed has been cumbersome, but there might be reasons to keep configuration updates manual (@pisuke ?) In which case we could have an additional set up step to copy the resources folder into the local site, but the `site_path` is configurable, so we wouldn't be able to have hard coded include's in the reference configuration which don't then have to be updated by the user. So maybe the answer is to make includes relative to the `site_path`? Not sure if it's actually an issue though or what the best approach is.

This is shown in the breakdown of compiled module_config which also identifies where items are declared or included:
[module_config breakdown.pdf](https://github.com/faucetsdn/daq/files/5792918/module_config.breakdown.pdf)

grafnu:
A few good points in there.... I'm trying to keep a somewhat abstracted
view of the system to separate out what it does vs. how it's used...
although they need to support each other. I think the one thing I didn't
have that I'll add in is making the includes relative. Right now it's fixed
relative to the daq install directory which doesn't make a lot of sense in
some places.  The other thing I'm thinking about is support for
variables in the include, e.g. $site_path/module_config.json -- but I'll
likely leave that to another day. I'm almost done with the relative include
fixup.. a lot more work than I expected (mostly in catching all the corner
test cases).

testing/test_modules.sh
```diff
@@ -36,5 +40,9 @@ cat $TEST_LIST | while read module args; do
     fi
 done
 
+echo TAPTAP
```
pbatta, line 43: TAPTAP?

grafnu:
Closing for now until I fix up the relative include bits.

testing/test_modules.sh
grafnu: That was some local debugging... removed now!

grafnu:
This has been updated with relative paths for include directives... PTAL.

pbatta (approved):

PR #734, opened 2021-01-12
daq changes for port up down events

henry54809:


daq/runner.py
```diff
@@ -360,6 +371,12 @@ def _handle_device_learn(self, vid, target_mac):
         else:
             device = self._devices.get(target_mac)
         device.dhcp_mode = DhcpMode.EXTERNAL
+        # For keeping track of remote port flap events
```
grafnu, line 374: blank line before comment?

etc/FORCH_VERSION
```diff
@@ -1 +1 @@
-0.48
+0.48.4
```
grafnu, line 1: 0.48.4 is a gmaster (internal only) tag .. should likely be 0.49  -- if you look at 0.48.3 it's a descendant of gmaster, not master.

henry54809:
Updated. PTAL

grafnu (approved):

PR #735, opened 2021-01-12
Support for vxlan

grafnu:


proto/system_config.proto
```diff
@@ -195,6 +195,9 @@ message SwitchSetup {
   // Interface name for external switch data plane
   string data_intf = 42;
 
+  // VXLAN configuration information
```
pbatta, line 198: Add this at the end, after the switch password config?

grafnu: done

pbatta (approved):

PR #741, opened 2021-01-13
Update internal config file filenames.

grafnu:


grafnu:
* base_config.json  -- defined in system.json as base_conf
* device_config.json -- keyed by MAC address
* type_config.json -- keyed by type field in device_config
* port_config.json -- keyed off of which switch port
* site_config.json -- as per site_path as configured in system.json

module_config.json is used as an output (merged from others) and consumed by modules.... no change there.

pbatta (approved):

PR #745, opened 2021-01-14
Return device result base on test results

didovesei:


daq/runner.py
```diff
@@ -870,6 +870,25 @@ def _combine_result_set(self, set_key, result_sets):
...
+            device_result = PortBehavior.failed
+        else:
+            device_result = self._calculate_device_result(test_results)
+
+        LOGGER.info('Sending device result for device %s: %s', mac, device_result)
+        self._device_result_client.send_device_result(mac, device_result)
+
+    def _calculate_device_result(self, test_results):
+        for module_name, module_result in test_results.get('modules', {}).items():
+            for test_name, test_result in module_result.get('tests', {}).items():
```
noursaidi, line 887: @grafnu  @didovesei I think this is right, but just flagging that this is different from the current report generation which compares against a 'required_result' from the module_config to determine if a test passes/fails. ( https://github.com/faucetsdn/daq/blob/34d14e92e4329138ca213de6f495ae3713b1b1c0/daq/report.py#L263  ) I can't find an instance where a test is expected to produce a "fail" result and looks like it was only in place for the info tests (which are being converted into normal pass/fail tests)

daq/host.py
```diff
@@ -437,7 +439,9 @@ def terminate(self, reason, trigger=True):
         self._monitor_cleanup()
         if self.target_port:
             self.runner.network.delete_mirror_interface(self.target_port)
-        self._finalize_report()
+
+        test_results = self._finalize_report()
```
henry54809, line 443: I think if you were to add an attribute “test_results” to the Device class and directly store the results on self.device, it would’ve made the runner class cleaner.

daq/runner.py
```diff
@@ -790,10 +791,6 @@ def _target_set_finalize(self, device, result_set, reason):
                 self._linger_exit = 1
         self._result_sets[device] = result_set
 
-        if self._device_result_client:
```
henry54809, line 793: This can be restored to replace line 771 and 814-816 when directly storing test_results on the device class; and self._send_device_results can be called with just 1 argument “device”.

daq/host.py
didovesei: So I see we access the host in runner.py through `runner.device.host`. And in host.py, we access the device thorugh  `host.device`. Are the two `device` the same object?

daq/runner.py
didovesei: I think the problem is that in my case the test_results is processed in `_target_set_cancel()`, which is called after `_target_set_finalize()`, which means when calling `_target_set_finalize()`, the test_result is not ready yet.

grafnu:
Can you be more specific about what you're tracking? I took a quick look at
the code and it (the code) is confusing me so I think there's something
"wrong" with it. I see the results processed in *both* locations, which
isn't quite what you're describing. I think the "right" answer would be to
have them (results) processed in target_set_finalize, thus avoiding the
problem you describe.  Either way I don't see "test_results" anywhere in
that file so not confident what you're referring to.

didovesei:
> I see the results processed in *both* locations, which isn't quite what you're describing. I think the "right" answer would be to have them (results) processed in target_set_finalize, thus avoiding the problem you describe.

So for the "results processed in both locations", they are different results. The `ReportGenerator` firstly processes the raw result, and generates the an in memory data, at the end of the test. What I did is to return that in-memory map to runner, so that it interprets that map (i.e., `test_results` in  `_target_set_cancel()`) to derive the `device_result` which will be sent to forch.

>  I think the "right" answer would be to have them (results) processed in target_set_finalize

So the problem here is that in my case, when calling target_set_finalize, the result is not generated by  `ReportGenerator` yet. The sequence in which the functions are called is: `Runner.target_set_complete -> Runner._target_set_finalize -> Runner._target_set_cancel -> target_host.terminate -> host._finalize_report()`

daq/runner.py
henry54809: You can modify where the the device.test_results are populated; for example, when there are no more tests remaining under that host. That way you have access to device.results under _target_set_finalize. Also the current logic where it returns PortBehavior.failed when any test module fails should also be left intact. 
Devices in host and in runner are the same objects; devices are passed around everywhere to reference information related to themselves.

didovesei: I checked the logic again and I don't think it is possible to move the report processing to _run_next_test() without chaning the way how the test is terminated in case of exception. I added some new logic on top of my changes to handle the exception case. PTAL..

daq/runner.py
```diff
@@ -884,9 +885,13 @@ def _send_device_result(self, mac, test_results):
 
     def _calculate_device_result(self, test_results):
         for module_result in test_results.get('modules', {}).values():
+            if report.ResultType.EXCEPTION in module_result:
```
henry54809, line 888: It should also check for non zero ResultType.RETURN_CODE

didovesei: Done

henry54809 (approved):

PR #748, opened 2021-01-17
Update UDMI tests

grafnu:


henry54809:
is the udmi version correct?

grafnu:
No, I need to finalize that still.

grafnu:
Updated, PTAL

henry54809 (approved):

PR #762, opened 2021-02-05
Feature/debian package files mapping

henry54809:


daq/configurator.py
```diff
@@ -81,8 +81,10 @@ def merge_config(self, base, config_file, traversed=None):
         traversed.add(config_file)
         if 'include' in config:
             include = config.pop('include')
-            self._log('Including config file %s' % include)
             including_config_file = os.path.join(os.path.dirname(config_file), include)
+            if include.startswith('/'):
```
grafnu, line 85: prefer non mutable variables:
```
including_config_file = os.path.join(os.path.dirname(config_file), include) if include.startswith('/') else include
```

daq/daq.py
```diff
@@ -19,7 +20,7 @@
 LOGGER = logger.get_logger('daq')
 ALT_LOG = logger.get_logger('mininet')
 
-_PID_FILE = 'inst/daq.pid'
+_PID_FILE = os.path.join(DAQ_RUN_DIR, 'daq.pid')
```
grafnu, line 23: Some places it's DAQ_RUN and others DAQ_RUN_DIR?  (e.g. bin/dump_network) -- any reason not to make them the same?

daq/gcp.py
```diff
@@ -5,14 +5,20 @@
...
 
-import firebase_admin
-from firebase_admin import credentials
-from firebase_admin import firestore
-
-from google.cloud import pubsub_v1
-from google.cloud import storage
-from google.cloud import logging
-from google.auth import _default as google_auth
+SUCCESSFUL_IMPORTS = True
```
grafnu, line 8: Move SUCCESSFUL_IMPORTS to the end of the try: block (again, non-mutated variables)

.github/workflows/tests.yml
```diff
@@ -15,7 +15,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        test: [base, many, aux, topo, modules, dhcp]
+        test: [aux]
```
grafnu, line 18: I assume this is debugging?

daq/daq.py
henry54809: So in DAQ I'm importing them as xxx_DIR and the env vars are the ones without DIR. I think DAQ_RUN, DAQ_LIB and DAQ_CONF by themselves are not very clear.

.github/workflows/tests.yml
henry54809: This was branched off my master which had this change. I'll clean this up when I move this out of draft.

daq/daq.py
grafnu: Got it. So, "internally" (python code) it's XXX_DIR, and then the user-facing surface is just XXX.  That's fine -- just wanted to make sure the user-facing surface was consistent.

henry54809:
PTAL.

bin/mudacl
```diff
@@ -3,31 +3,31 @@
...
+mud_files=$DAQ_LIB/mud_files/
+acl_templates=$DAQ_RUN/acl_templates
+port_acls=$DAQ_RUN/port_acls
 
 cd $ROOT
-mudacl/bin/test.sh
+$DAQ_LIB/mudacl/bin/test.sh
 
-source etc/config_base.sh
+source $DAQ_LIB/etc/config_base.sh || source $DAQ_LIB/bin/config_Base.sh
```
grafnu, line 16: Why "config_Base.sh"?

etc/PACKAGE_FILES_MAPPING
```diff
@@ -0,0 +1,3 @@
+export DAQ_LIB=/usr/lib/daq
```
grafnu, line 1: Maybe put this in debian/FILES_MAPPING?    I'd prefer not to have "FILES_MAPPING" and "PACKAGE_FILES_MAPPING" in the same place without qualifying FILES_MAPPING (e.g. SETUP_FILES_MAPPING rather than FILES_MAPPING).  Either way (either different directories or two unique prefixes) I think would work.

(Actually, if they're in the same place I'd recommend a suffix not a prefix so they're sorted together, e.g. FILES_MAPPING_SETUP and FILES_MAPPING_DEBIAN or something...)

henry54809:
Updated; PTAL.

grafnu (approved):

PR #766, opened 2021-02-12
parallelize check, cleanup config, better fot logging

grafnu:


daq/runner.py
```diff
@@ -926,18 +926,26 @@ def _send_device_result(self, mac, test_results):
...
+                LOGGER.warning('Failing report due to module exception')
+                failed = True
+
+            return_code = module_result.get(report.ResultType.RETURN_CODE)
+            if return_code:
+                LOGGER.warning('Failing report due to module return code %s', return_code)
+                failed = True
+
+            module_tests = module_result.get('tests', {})
+            for test_name in module_tests:
```
didovesei, line 941: Maybe `for test_name, test_result  in module_tests`.items() is better?

daq/runner.py
```diff
@@ -926,18 +926,26 @@ def _send_device_result(self, mac, test_results):
         self._device_result_client.send_device_result(mac, device_result)
 
     def _calculate_device_result(self, test_results):
+        failed = False
```
didovesei, line 929: I am not sure why we always check all the three items: exception, return code, and module test results, since one failure is already enough to indicate that the test has failed. Is it because we want to have a more verbose logging on the test results?

grafnu:
Yes -- I wanted more comprehensive logging; that was the main reason for
the change.  There's no real functional difference.

cmd/build
```diff
@@ -25,7 +25,6 @@ DOCKER_IMAGE_LIST=docker_images.txt
 DOCKER_IMAGE_VER=docker_images.ver
 
 cd $ROOT
-source etc/FILES_MAPPING
```
henry54809, line 28: Why are these lines removed?

etc/config_base.sh
```diff
@@ -1,5 +1,7 @@
 # Script file included by all setup scripts to load local config.
 
+source etc/FILES_MAPPING
```
henry54809, line 3: Putting this here won’t work with the Debian package files locations since this script is also run and the env vars will not be the right ones.

grafnu:
The goal was to consolidate it into the shared configuration mechanism --
since it's a form of configuration. Trying to maintain the abstraction and
not bleed "etc/FILES_MAPPING" everywhere.  Seems like debian might be a
special case, but we should sort through that separately.

grafnu:
OK -- I"ll have a look more closely at the debian packages. I saw the
explicit reference to debian/FILES_MAPPING and probably just made a bad
assumption about how it worked. I think the right solution (if I can figure
it out) is to make sure that the FILES_MAPPING that's installed with debian
is the right one (or doesn't exist) so that it doesn't cause problems.

henry54809:
How about in setup.py, replace the line "source etc/FILES_MAPPING" in etc/config_base.sh with the content of debian/FILES_MAPPING?
The problem I was encountering was it is not known if DAQ is in a package form in scripts without the env vars.

didovesei (approved):

grafnu:
Ok, I updated and tested a solution that works with debian. Basically it only includes the file mappings if they're not already defined (which is the case with debian). I also moved the common config thing to bin/ (from etc/) -- a bit cleaner since then it's not special cased.

grafnu:
PTAL (mainly Haoli for the debian part)

henry54809 (approved):

PR #773, opened 2021-02-16
Exception Handling in target_set_cancel and grpc timeout 

henry54809:


daq/runner.py
```diff
@@ -211,7 +211,10 @@ def _init_daq_run_id(self):
     def _init_device_result_client(self):
         server_port = self.config.get('device_reporting', {}).get('server_port')
         if server_port:
-            return DeviceReportClient(server_port=server_port)
+            timeout = self.config['device_reporting'].get('rpc_timeout_sec')
+            if timeout:
+                return DeviceReportClient(server_port=server_port, rpc_timeout_sec=timeout)
```
grafnu, line 216: NB: I'm surprised Python doesn't have a more elegant way of doing this!  I've run into the same problem before...

proto/system_config.proto
```diff
@@ -253,4 +253,5 @@ message Interface {
 message DeviceReporting {
   // Server port
   int32 server_port = 1;
+  int32 rpc_timeout_sec = 2;
```
grafnu, line 256: add comment

daq/device_report_client.py
```diff
@@ -28,15 +31,28 @@ def send_device_result(self, mac, device_result):
...
         }
-        for port_event in self._stub.GetPortState(dict_proto(device, Device)):
+        result_generator = self._stub.GetPortState(dict_proto(device, Device))
+        def cancel_request():
+            result_generator.cancel()
+
+        self._pending_requests.append(cancel_request)
+        for port_event in result_generator:
             callback(port_event)
+        if cancel_request in self._pending_requests:
```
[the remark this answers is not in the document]
henry54809: Originally I had a different place where handlers would be removed.

henry54809:
PTAL.

proto/system_config.proto
```diff
@@ -253,5 +253,6 @@ message Interface {
 message DeviceReporting {
   // Server port
   int32 server_port = 1;
+  // GRPC timeout in seconds
```
grafnu, line 256: nit: blank line before

grafnu (approved):

PR #774, opened 2021-02-17
use abs path for files mapping

henry54809:


etc/FILES_MAPPING
```diff
@@ -1,3 +1,4 @@
-export DAQ_LIB=.
+ROOT=$(realpath $(dirname $BASH_SOURCE)/..)
+export DAQ_LIB=$ROOT
```
grafnu, line 2: Don't all of these need to be adjusted, not just DAQ_LIB?  Or I guess the others are always used relative to the run directory?

grafnu (approved):

etc/FILES_MAPPING
henry54809: we can but they don't have to be adjusted. The other 2 are always relative to the daq root.

PR #777, opened 2021-02-22
Reference & CI configuration Files Revamp

noursaidi:
- Updates the reference configuration file names to match new structure
- Remove the reference to CI/faux specific configuration in reference files (faux signing certificate, faux password list)
- Update reference configuration files to use the new include functions, and eliminate all repetition of tests or information.
- General test clean up - non-existent tests removed, Information test category removed
- Update the test_site configuration to use the reference configuration files,  and the new structure. This should eliminate a repeat of the issue where we had three different sets of configuration files as only one set should be updated


resources/setups/qualification/qualifcation_tests.json
```diff
@@ -0,0 +1,4 @@
+{
+  "include": "../common/common_tests.json"
```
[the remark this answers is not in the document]
noursaidi: Agree, renamed - make more sense with that name

resources/setups/qualification/site_config.json
```diff
@@ -8,7 +8,7 @@
     "categories": [ "Base", "Connection", "Security", "NTP", "DNS", "Communication", "Protocol", "PoE", "IoT"],
     "expected": [ "Required Pass", "Required Pass for PoE Devices", "Required Pass for BACnet Devices", "Required Pass for IoT Devices", "Recommended Pass" ]
   },
-  "include": "../qualification/qualification_tests.json",
+  "include": "../../resources/setups/qualification/qualification_tests.json",
```
grafnu, line 11: name is redundant, something is wrong here.  Should this be qualification/test_config.json?

The goal is to have the actual filename (.json) be semantically meaningful to describe what kind of information is in that file (test config?), while the directory path indicates which one (for qualification).

noursaidi: Agree as per your previous comment on the `common/test_config.json` would be the suitable name

Thinking more, what I've done here may not appropriate or needed yet. My intent was to try and avoid repetition of tests by including from a `common/test_config.json`, but there may be valid reasons to categorise tests differently in different test environments which is what is in these files. Maybe best to have only a `baseline/test_config.json` which describes all the test for now.

grafnu:
Actually, if you could make it "tests_config.json" (plural) that would be
good. "test_config.json" is unfortunate since it's easily confused with a
not-real "test" config. :-/

I could go either way on your other point (common vs. baseline)... It
doesn't really matter now as long as there is *one* place to factor out any
common parameters.  What's bad is if there's duplication of things, because
that's then more places to change it in the future. As long as there's one
place it's much easier to change down the road as we evolve things.

grafnu (approved):

PR #786, opened 2021-02-27
fixes for FOT

henry54809:
1. Fix for testing utils files mapping
2. Change dhcp match regex to support newer versions of tcpdump
3. Adding connection checks for faucet socket and attempts to reconnect if connection is dropped. 
4. Adding integration tests for testing faucet socket reconnections. 
4. Adding system config for specifying faucet event timeouts. 
5. Adding default FAUCET_EVENT_SOCK_HEARTBEAT env for starting faucet. 
6. Adding separate queue for storing debounces port events. Events in this queue will be processed before other faucet events.
7. Fix duplicate remote device triggering.


daq/faucet_event_client.py
```diff
@@ -11,19 +12,25 @@
 
 LOGGER = logger.get_logger('fevent')
 
+
 class FaucetEventClient():
     """A general client interface to the FAUCET event API"""
 
     FAUCET_RETRIES = 10
     _PORT_DEBOUNCE_SEC = 5
+    _EVENT_TIMEOUT_SEC = 10
```
grafnu, line 21: _DEFAULT_

daq/faucet_event_client.py
```diff
@@ -65,12 +69,25 @@ def has_event(self, blocking=False):
             if '\n' in self.buffer:
                 return True
             if blocking or self.has_data():
-                data = self.sock.recv(1024).decode('utf-8')
-                with self._buffer_lock:
-                    self.buffer += data
-            else:
+                data = self.sock.recv(2048).decode('utf-8')
```
grafnu, line 72: 2048 -> NAMED_CONSTANT

daq/faucet_event_client.py
```diff
@@ -65,12 +69,25 @@ def has_event(self, blocking=False):
...
+                    self._last_message = time.time()
+                    with self._buffer_lock:
+                        self.buffer += data
+            if not blocking and self._check_socket_connection():
                 return False
 
+    def _check_socket_connection(self):
+        time_elapsed = time.time() - self._last_message
+        if self._event_timeout_sec and time_elapsed >= self._event_timeout_sec:
+            LOGGER.info('Attempting to reconnect to faucet socket %s ' %
```
grafnu, line 83: move comment to connect() or ensure "os.getenv('FAUCET_EVENT_SOCK')" is only called in one place

daq/faucet_event_client.py
```diff
@@ -65,12 +69,25 @@ def has_event(self, blocking=False):
...
+                        os.getenv('FAUCET_EVENT_SOCK'))
+            time.sleep(self._event_timeout_sec)  # Ensure faucet socket's been disconnected.
+            self.connect()
+            time.sleep(self._event_timeout_sec)  # Ensure a new heart beat has happened.
```
grafnu, line 87: seems like this sleep should be in connect() itself.  Why is "reconnect" behavior different than initial connection?

daq/runner.py
```diff
@@ -253,6 +256,8 @@ def initialize(self):
         LOGGER.debug('Attaching event channel...')
         self.faucet_events = faucet_event_client.FaucetEventClient(self.config)
         self.faucet_events.connect()
+        # sock fileno will change if the socket is ever reconnected.
```
grafnu, line 259: I don't know what to make of this comment. I understand what it's saying, but I don't know why it's here.  If it's true (it will change), then doesn't something need to re-call this function to get the new value?

It doesn't seem like this class variable (faucet_events_fd) is very useful, why does it need to exist at all?  It seems like it's introducing a problem of inconsistent state?

How does the stream monitor (called through self.monitor_stream) pick up the new file descriptor?

daq/stream_monitor.py
```diff
@@ -141,11 +142,11 @@ def event_loop(self):
         try:
             if not fds and self.idle_handler:
                 self.idle_handler()
-                # Check corner case when idle_handler removes all callbacks.
-                if not self.callbacks:
-                    return False
             if self.loop_hook:
                 self.loop_hook()
+                # Check the corner case when loop_hook removes all callbacks.
```
grafnu, line 147: why is the idle_handler check going away?  Seems like if anything it would need to be in both places?

proto/system_config.proto
```diff
@@ -138,6 +138,9 @@ message DaqConfig {
 
   // Path to include file
   string include = 56;
+
+  // The maximum time in seconds between messages from faucet
+  int32 faucet_event_timeout_sec = 57;
```
grafnu, line 143: I'd like to keep the concept of "faucet" out of the exposed config, so maybe just call this "event_timeout_sec"

daq/faucet_event_client.py
```diff
@@ -65,12 +69,25 @@ def has_event(self, blocking=False):
...
+            if not blocking and self._check_socket_connection():
                 return False
 
+    def _check_socket_connection(self):
+        time_elapsed = time.time() - self._last_message
+        if self._event_timeout_sec and time_elapsed >= self._event_timeout_sec:
+            LOGGER.info('Attempting to reconnect to faucet socket %s ' %
+                        os.getenv('FAUCET_EVENT_SOCK'))
+            time.sleep(self._event_timeout_sec)  # Ensure faucet socket's been disconnected.
+            self.connect()
```
grafnu, line 86: as per the comments in runner.py, I think some kind of notification will be required for when the socket changes so that it can be properly monitored. Currently, it seems like the socket can change, creating a new fd, but it won't be properly monitored by the calling code -- how does this change even work as it is?

daq/faucet_event_client.py
henry54809: It was originally 1024. I changed it to 2048 to test some things but forgot to change back. Do you still think this should be a named constant? This constant isn’t very useful.

daq/faucet_event_client.py
henry54809: It needs to wait that much time to ensure any previous client connections are closed on the faucet end. There maybe other reasons for the disconnection but under first connect, there is no reason to expect a client is already connected by default.

daq/runner.py
henry54809: The fd shouldn’t change since the old fd can still be polled as expected. I don’t know why there’s a new one but I can modify the steam monitors reconnect to overwrite the   fd in callbacks dict so this doesn’t need to be there.

daq/stream_monitor.py
henry54809: Unless I’m missing something, I think only shutdown can remove all callbacks. Since the shutdown is moved out of idle handler to loop hook, I moved the check to be under loop hook.

proto/system_config.proto
henry54809: I thought about it but if I were to leave out faucet, this configuration is overly vague. It’s like renaming faucet_event_client to just event_client

daq/runner.py
henry54809: I found a much simpler way to handle sock fd changes. PTAL my newest commit.

grafnu:
What's the theory of operation?  How does that work when stream_monitor
blocks?

henry54809:
Stream monitor never blocks since it always has a poll timeout. Faucet events are processed separately for every loop since the heartbeat events will always be there. The faucet_event_client doesn’t use the fd for polling; it directly uses the socket object which is updated on reconnection. 
 
> What's the theory of operation? How does that work when stream_monitor blocks?
> […](#)

grafnu:
Right, ok -- I'm sure that will work -- but I don't think it's the most
elegant solution, but I suppose that "issue" is already there since that
behavior is assumed for the heartbeat to function. NB: Stream monitor
does block, just not indefinitely (there's a technical difference).   Maybe
just add a brief one-line comment about the assumption (that stream_monitor
has a poll timeout).

proto/system_config.proto
grafnu: The configuration is an external interface specification, and really shouldn't be bleeding the fact that "faucet" is used internally.  This is different than the code (faucet_event_client) since that's internal to the system (implementation not interface). I'm more concerned about bleeding the abstraction than I am it being overly vague. The concept of an "event timeout" has meaning even when it's not faucet (e.g. could map to whatever system it's assumed to be getting such events from).

daq/stream_monitor.py
grafnu: It sounds like you're using a-priori knowledge about some other class as applied to this class -- a dangerous game.  In theory, the idle_handle could remove all the callbacks, there's nothing that prevents it from doing so.  You're statement should be "currently only the shutdown method from some other class actually removed all callbacks" -- anything _can_ do it, which is different than what does it now.

Let me put it this way:  Is there something wrong with the check after idle_handler for this, or were you removing it because you knew it wasn't used?

grafnu (commented):
Still a few unresolved comments.

PR #789, opened 2021-03-04
Handle faucet event socket disconnects

henry54809:


daq/runner.py
```diff
@@ -290,8 +290,21 @@ def _handle_faucet_events_locked(self):
...
     def _handle_faucet_events(self):
+        event = None
         while self.faucet_events:
-            event = self.faucet_events.next_event()
+            try:
+                event = self.faucet_events.next_event()
+            except DisconnectedException:
+                self.monitor_forget(self.faucet_events.sock)
+                self.faucet_events.connect()
+                self.monitor_stream('faucet', self.faucet_events.sock,
```
grafnu, line 300: can you refactor this to combine this call with the other one (in main_loop) -- there should just be one place in the code that this is instantiated (e.g., the constants 'faucet' and 10 should only be in one place).  Ideally you could factor out the connect as well, since that's the essence of the operation.  Unfortunately the other calls to "connect" and "monitor_stream" are currently in different places (initialize and main_loop), but ideally those would be refactored as well -- but don't want this to turn into a cascading PR...

grafnu (commented):
Overall much more elegant solution, I like it.  Just one comment about refactoring some of the common code...

grafnu (approved):

PR #790, opened 2021-03-11
Base commit for authenticator module

anurag6:
TODO:
- Add support for timeouts
- Add config file to configure radius and eap socket params

REVIEW GUIDE:
Files that mostly involve packet packing, parsing and structure(heavily borrowed from chewie with some interfacing changes and removal of extras) and hence, can be ignored:
- auth_821x.py
- eap.py
- ethernet_packet.py
- mac_address.py
- message_parser.py
- radius.py
- radius_attributes.py
- radius_datatypes.py

Files under etc/ are mostly freeradius and wpa supplicant related and can be ignored.

Files to focus on:
- authenticator.py
- eap_module.py
- eap_socket.py
- radius_scoket.py
- radius_module.py

Optional:
- docker-compose.yaml
- utils.py


grafnu (commented):
* Isn't the wpasupplicant part of the DUT?  If so, all files associated with that should go under the docker/ top-level directory as they're not part of the test (rather go into the faux model)
* The module needs to be one docker file, not separated out. So, the authenticator and freeradius pieces need to be combined/consolidated.

anurag6:
MAkes sense. Filed a separate [PR](https://github.com/faucetsdn/daq/pull/791) for freeradius and wpa_supplicant stuff

subset/dot1x/Dockerfile.authenticator
```diff
@@ -0,0 +1,13 @@
+FROM ubuntu:18.04
```
grafnu, line 1: This should be based on aardvark or aardvark2, helps to amortize the build cost.

anurag6: Makes sense. Changed.

grafnu (approved):

PR #791, opened 2021-03-12
Add freeradius and wpasupplciant files

anurag6:
A separate PR for freeradious and wpasupplicant files to improve readability of https://github.com/faucetsdn/daq/pull/790


grafnu (commented):
Your file placement doesn't match your diagram https://docs.google.com/drawings/d/1KCFvy85tlysf6F0f2eVpQxL3Lr8f1hM-AL0RID-d3XU/edit

If it ends up in the "authenticator test container" then it goes in subset/ somewhere, if it it's in faux or gateway then it goes in docker/ somewhere.

etc/freeradius/certs/01.pem
```diff
@@ -0,0 +1,80 @@
+Certificate:
```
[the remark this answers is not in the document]
anurag6: Got it. Changed and merged wpasupplciant with faux1

docker/modules/Dockerfile.faux1
```diff
@@ -8,7 +8,7 @@ FROM daqf/aardvark:latest as java_build
 # Run this separately so it can be shared with other builds.
 RUN $AG update && $AG install openjdk-8-jre
 
-RUN $AG update && $AG install openjdk-8-jdk git
+RUN $AG update && $AG install openjdk-8-jdk git vim iproute2
```
grafnu, line 11: take a closer look at the structure of the docker file -- you're messing up the shared java build image by putting this here and it won't end up in the image you think it will.

docker/modules/Dockerfile.faux1
```diff
@@ -30,6 +30,12 @@ RUN sh ssh_build.sh
     
 FROM daqf/aardvark:latest
 
+# wpa_suplicant for dot1x
+
+RUN $AG update && $AG install wpasupplicant
```
grafnu, line 35: this too -- in the wrong spot in this file.  Look at the "FROM" directives and see how they're used, and then how docker layers are shared across builds.

anurag6: Yup. My bad. Changed.

grafnu (approved):

PR #792, opened 2021-03-17
Set up Gauge

didovesei:


proto/system_config.proto
```diff
@@ -183,6 +183,9 @@ message SwitchSetup {
   // Local port for an alternate OpenFlow controller
   int32 alt_port = 16;
 
+  // Gauge OpenFlow Port
+  int32 gauge_port = 15;
```
grafnu, line 187: this shouldn't expose the concept of "gauge" here, as that's an internal implementation detail.  How is this functionally different than lo_port? or alt_port?  Doesn't the switch just get programmed with two controllers? So would that be { lo_port, lo_port2 }?

didovesei: `lo_port_2` LTGM. Just wanted the user to be able to configure it in DAQ... Will update accordingly.

didovesei: Changed to `lo_port_2`. PTAL..

grafnu (approved):

PR #794, opened 2021-03-18
Run wpasupplicant in faux

anurag6:


docker/include/bin/start_faux
```diff
@@ -277,5 +277,8 @@ else
     echo Runtime config $conf_file not found.
 fi
 
+# Trigger wpasupplicant
```
grafnu, line 280: this should be configurable -- see all the places where it says ${options -- I'd just use wpa -- so ${options[wpa]} should suffice.  (Also -- comment isn't helpful -- it's inherently obvious what the code does, so it's just extra words!)

docker/modules/Dockerfile.faux1
```diff
@@ -83,6 +83,7 @@ RUN mv /usr/sbin/tcpdump /usr/bin/tcpdump
 
 RUN $AG update && $AG install wpasupplicant vim iproute2
 
-COPY docker/include/etc/wpasupplicant/ /etc/wpasupplicant/
+COPY docker/include/etc/wpasupplicant/ /tmp/wpasupplicant/
+COPY docker/include/etc/wpasupplicant/cert/ /tmp/cert/
```
grafnu, line 87: this seems redundant -- why do you need this if you're copying the entire directory already?

grafnu (approved):

docker/modules/Dockerfile.faux1
anurag6: Yeah, that was because of the way the .conf file was configured. Changed.

docker/include/bin/start_faux
anurag6: Changed.

docker/include/bin/start_faux
```diff
@@ -277,5 +277,9 @@ else
     echo Runtime config $conf_file not found.
 fi
 
+if [ -n "${options[wpa_supplicant]}" ]; then
```
grafnu, line 280: please use wpa like my original comment -- things become unwieldy very quickly with long names for these options.

Also note that you'll have to add that option in wherever you launch the faux container -- not sure if that's landed in a PR yet or not (doesn't need to be included here -- that's just FYI).

anurag6: Changed.

grafnu (approved):

PR #795, opened 2021-03-18
Add freeradius files necessary for dot1x tests

anurag6:
Files under subset/dot1x/freeradius/ can be ignored


subset/dot1x/test_dot1x
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash -e
 
+freeradius -fxx -l stdout &
```
grafnu, line 3: how does this work running this in the background without the while loop?  Seems like it would run this and then just go all the way through to the report without ever waiting for anything?

anurag6: It just starts the freeradius server before calling the python script. The idea is to make sure test_dot1x.py blocks until the authenticator module returns a definitive result and then write to the file. For now, it just goes through since test_dot1x.py writes to the file directly.

grafnu:
At least add a todo explaining what is missing (e.g. "this will always fail
until...")

subset/dot1x/test_dot1x
anurag6: Trevor, it just starts a RADIUS server. The fail/pass would be handled by the python call. Since it is currently just something that writes directly, I think it should be fine.

grafnu:
You're missing the point.  Anything submitted, even if partial, should be
logically consistent. It's not about this line "starting a radius server"
-- that's obvious -- it's that the other bit (the python code) isn't quite
there yet and so everything will be failing.

The criteria is that you should be able to "walk away" from anything
submitted and have it make logical sense.

It's just general good practice -- yes, this is a trivial case, but it
should make it that much easier to add a TODO that marks what it *should*
be doing even though it's not.

subset/dot1x/test_dot1x
anurag6: Added a todo to test_dot1x.py

grafnu (approved):

PR #801, opened 2021-03-24
Acl changes for EAP

anurag6:


daq/topology.py
```diff
@@ -426,6 +426,19 @@ def _generate_main_acls(self):
         local_acl = []
         acls = {}
 
+        dot1x_pri_ports = []
+
+        for devices in self._set_devices.values():
+            for device in devices:
+                dot1x_pri_ports.extend(device.gateway.get_possible_test_ports())
+
+        dot1x_pri_rule = self._ethertype_to_port_acl(0x888e, dot1x_pri_ports)
```
grafnu: create constant for 0x888e

daq/topology.py
```diff
@@ -531,9 +550,10 @@ def _generate_port_acl(self, port):
         if target_mac:
             assert self._append_acl_template(rules, 'baseline'), 'Missing ACL template baseline'
             self._append_device_default_allow(rules, target_mac)
-            self._write_port_acl(port, rules, file_path)
         else:
-            self._write_port_acl(port, self._make_default_acl_rules(), file_path)
+            rules = self._make_default_acl_rules()
```
grafnu, line 554: don't assign directly to rules -- append?

daq/topology.py
```diff
@@ -630,6 +650,13 @@ def _append_device_default_allow(self, rules, target_mac):
             acl = {'rule': subrule}
             self._append_augmented_rule(rules, acl)
 
+    def _ethertype_to_port_acl(self, ether_type, ports):
```
grafnu, line 653: can you add eth_type to _make_acl_rule and use that instead?   This function is essentially doing the same work that you're doing here... ?

daq/topology.py
```diff
@@ -519,6 +534,10 @@ def _generate_port_acl(self, port):
         target_mac = None
         rules = []
 
+        if port != self.sec_port:
```
grafnu, line 537: why is this check necessary?

daq/topology.py
anurag6: I just didn't want to change the functionality of what it was doing. I assume we send the appended rules only for the 'if' case.

daq/topology.py
anurag6: Since the ACL is applied to the sec_port, no point trying to add a rule which diverts it back.

grafnu: But this function should never be called with port == self.sec_port
If it is, then that's the core issue.  If you're not sure, then add an assert to verify.

daq/topology.py
grafnu: In this case I'd assume it was an oversight in the original code, and just append the rules (keeps the overall pattern of the function more consistent).  If it's a problem (will fail some test somewhere), then we can re-visit.

daq/topology.py
anurag6: Changed.

daq/topology.py
anurag6: Changed.

daq/topology.py
anurag6: Changed.

daq/topology.py
anurag6: Changed.

daq/topology.py
```diff
@@ -426,18 +427,16 @@ def _generate_main_acls(self):
...
+                dot1x_ports.extend(device.gateway.get_possible_test_ports())
 
-        dot1x_pri_rule = self._ethertype_to_port_acl(0x888e, dot1x_pri_ports)
-        incoming_acl.append(dot1x_pri_rule)
+        self._add_acl_rule(incoming_acl, eth_type=self._DOT1X_ETH_TYPE, ports=dot1x_ports)
 
-        dot1x_sec_ports = list(range(1, self.sec_port))
-        dot1x_sec_rule = self._ethertype_to_port_acl(0x888e, dot1x_sec_ports)
-        secondary_acl.append(dot1x_sec_rule)
+        dot1x_ports = list(range(1, self.sec_port))
```
grafnu, line 438: don't reuse variable name

anurag6: Changed.

grafnu (approved):

PR #812, opened 2021-03-31
Add ACL State Collector

didovesei:
Implementing the AclStateCollector module as part of the big traffic analyzer component.


daq/acl_state_collector.py
```diff
@@ -0,0 +1,70 @@
...
+    def __init__(self):
+        self._switch_configs = {}
+
+    def get_port_acl_count(self, switch, port, port_acl_samples):
+        """Return the ACL count for a port"""
+
+        switch_config = self._switch_configs.get(switch)
+        if not switch_config:
+            LOGGER.warning('Switch not defined in Faucet dps config: %s', switch)
+            return {}
```
grafnu, line 22: rather than return empty, I think there should be a mechanism that includes an error indicator in the result. All of the error cases here seem fairly fundamental and shouldn't happen in normal use, correct?

daq/acl_state_collector.py
```diff
@@ -0,0 +1,70 @@
...
+
+        acls_config = port_config.acls_in
+        if not acls_config:
+            LOGGER.warning('No ACLs applied to port: %s:%s', switch, port)
+            return {}
+
+        return self._get_port_rules_count(switch, port, acls_config[0], port_acl_samples)
+
+    # pylint: disable=protected-access
+    def _get_port_rules_count(self, switch, port, acl_config, acl_samples):
```
grafnu, line 37: is acl_samples == port_acl samples?  Unless there's a good reason, use consistent variable name.

daq/acl_state_collector.py
```diff
@@ -0,0 +1,70 @@
...
+                    continue
+                if int(sample.labels.get('in_port')) != port:
+                    continue
+                rule_map = rules_map.setdefault(rule_config['description'], {})
+                rule_map['packet_count'] = int(sample.value)
+                has_sample = True
+                break
+
+            if not has_sample:
+                LOGGER.warning(
```
grafnu, line 62: likewise, should there be a specific error indicator returned in the result?

daq/acl_state_collector.py
```diff
@@ -0,0 +1,70 @@
+"""Process ACL states"""
+
+from __future__ import absolute_import
+
+import logger
+
+LOGGER = logger.get_logger('acl')
```
grafnu, line 7: aclstate?  I think "acl" might be too short to be "memorable"...

daq/acl_state_collector.py
didovesei: Updated. Also added protobuf message to define the output of the function. The `expected_acl_count` in test_acl_state_collector gives an idea about what the output looks like. PTAL..

daq/acl_state_collector.py
didovesei: Updated so that acl_samples is used in all the places.

daq/acl_state_collector.py
didovesei: Done

daq/acl_state_collector.py
didovesei: updated

daq/acl_state_collector.py
```diff
@@ -59,11 +55,40 @@ def _get_port_rules_count(self, switch, port, acl_config, acl_samples):
...
+
+        acls_config = port_config.acls_in
+        if not acls_config:
+            error = f'No ACLs applied to port: {switch}, {port}'
+            LOGGER.error(error)
+            error_list.append(error)
+            return None, error_map
 
-        return {'rules': rules_map}
+        return acls_config[0], None
```
grafnu, line 91: also check if len(acls_config) != 1 -- since if it's >1 then information will be lost which is likely confusing -- or, if that's not possible do an assert to check len == 1

proto/acl_counts.proto
```diff
@@ -0,0 +1,29 @@
...
+ */
+message AclCounts {
+  // List of ACL counts indexed by device MAC addresses
+  map<string, AclCount> devices = 1;
+}
+
+/*
+ *  ACL counts for a device
+ */
+message AclCount {
```
grafnu, line 17: I think this should be RuleCounts. AclCounts == map -> AclCount, RuleCounts --> map -> RuleCount

grafnu (approved):
Overall looks fine -- just some minor error conditions to check and slight name change

proto/acl_counts.proto
didovesei: So something like this?

```
message AclCounts {
  // List of ACL counts indexed by device MAC addresses
  map<string, RuleCounts> devices = 1;     <--------------------- changed from AclCount to RuleCounts
}

message RuleCounts {            <--------------------------- changed from AclCount to RuleCounts
  // List of rule counts indexed by the descriptions of the rules
  map<string, RuleCount> rules = 1;
  repeated string errors = 2;
}
```

grafnu:
I think "RuleCounts" is fine -- since it's a collection of rules (key) that
point to individual RuleCount messages (which contain the counts)

AclCounts doesn't seem to make sense... It's a collection of Devices (key)
that points to RuleCounts -- nothing specifically to do with ACLs, and
doesn't follow the pattern of the other message (key and message name are
related).

PR #814, opened 2021-04-01
Use installed python version as default python

didovesei:
It seems that Mininet has been compiled with Python 2 instead of Python 3. What happened is that Python3 is installed as `/opt/hostedtoolcache/Python/3.8.8/x64/bin/python`. When python3 is installed, the target of `/usr/bin/python` is not updated accordingly and still points to python2.7.  And when compiling mininet, the script runs python with sudo. As `/opt/hostedtoolcache/...` is not included in the root user's PATH, it will use `/usr/bin/python`, i.e., python2, as the interpreter.

I found this because there is an issue in my own repo's github action: the mininet compilation failed because the mininet installation script installs setuptools in python3 but uses python2 for compiling. 

https://github.com/didovesei/daq/runs/2243711116?check_suite_focus=true

Seems that the complilation failure arises only in github action image 20210330.1, while faucetsdn's image is not updated yet.

### DIFF
--- a/bin/setup_dev
+++ b/bin/setup_dev
@@ -58,6 +58,11 @@ $AG install \
     python$PVERSION python3-pkg-resources python3-setuptools \
     python$PVERSION-dev python3-pip python emacs-nox python$PVERSION-venv nmap
 
+readlink -f $(which python)
+PYTHON_BIN=$(readlink -f $(which python))
+echo Python binary is at $PYTHON_BIN
+sudo ls -fs $PYTHON_BIN /usr/bin/python
+
 if [ -d mininet ]; then
     echo Checking mininet version matches $MININETV...
     targetrev=$(cd mininet; git rev-parse $MININETV)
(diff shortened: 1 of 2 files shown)


bin/setup_dev
```diff
@@ -58,6 +58,11 @@ $AG install \
     python$PVERSION python3-pkg-resources python3-setuptools \
     python$PVERSION-dev python3-pip python emacs-nox python$PVERSION-venv nmap
 
+PYTHON_BIN=$(readlink -f $(which python))
+echo Python binary is at $PYTHON_BIN
+sudo ln -fs $PYTHON_BIN /usr/bin/python
```
grafnu, line 63: this is dangerous and really not a good idea -- you're assuming that this is the only thing that will ever run on the target system, which is not a good idea... and could seriously mess up the platform otherwise.

What scope is this meant to cover?  If it's for just testing then it could be something specific in the github actions page.  

If it's really something with mininet, then we might be able to push/make a change to that package to do the needful (e.g. call python3 directly), or allow somehow running in a venv or with a specific python version.

Another alternative would be to put something in the binhack/ directory and add that to the PATH environment... but that wouldn't work if the mininet executable is being run sudo (which resets the path).

grafnu (changes requested):

grafnu (approved):

PR #815, opened 2021-04-01
Make hold tests run natively

henry54809:
Need to make this change so I can force a sequestering timeout on forch's FOT testing. I also removed the concept of core tests in daq/host.py since I think this brings more confusion than benefits.


testing/test_base.sh
```diff
@@ -62,12 +62,20 @@ cat inst/result.log | tee -a $TEST_RESULTS
...
-DAQ_TARGETS=test_hold cmd/build
+
+echo %%%%%%%%%%%%%%%%%%%%%% Missing Docker Image | tee -a $TEST_RESULTS
+# Check that an open port causes the appropriate failure.
+docker rmi daqf/test_bacnet:latest # Check case of missing image
+cmd/run -s 
+echo DAQ result code $? | tee -a $TEST_RESULTS
+cat inst/result.log | tee -a $TEST_RESULTS
+
+if [ -z "$CI" ]; then
```
grafnu, line 77: why is this check necessary?  I'd be hoping the testing/ script would behave deterministically -- why does this only need to run outside of testing?

daq/host.py
```diff
@@ -176,9 +175,8 @@ def _get_test_config(self, test):
         return get_test_config(self._loaded_config, test)
 
     def _test_enabled(self, test):
-        fallback_config = {'enabled': test in self._CORE_TESTS}
-        test_config = self._get_test_config(test) or fallback_config
-        return test_config.get('enabled', True)
+        test_config = self._get_test_config(test) or {}
```
grafnu, line 178: what's the impetus behind this change?  The intent was to have a basic set of things that would do something even in the absence of a valid external configuration.

subset/hold/test_hold
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash -e
```
grafnu, line 1: should the old (docker/include/bin) test_hold be deleted?

testing/test_base.sh
henry54809: It can save some time since the image is not needed afterwards

subset/hold/test_hold
henry54809: It's used in building of another container.

grafnu:
If not needed then why restore it at all?  A more clean alternative might
be to save the image # of the previously tagged image, and then just re-tag
it. Then behavior is consistent in all cases and very fast.

grafnu:
Ok -- I think then we should change the name of the other file because in
the long run it'll be ultimately more confusing to have two "test_hold"
files in different places. Maybe rename the other one to "null_test" or
something like that?  Or "block_forever"?  (Semantic name based on what it
does, since it's not the "hold" test anymore.)

daq/host.py
henry54809: This brings inconsistencies in which tests are enabled; some test(as in the core tests) will run when "enabled" is not in the config and some tests won't. I have found this core concept + host_tests + module_configs to bring extra confusions on which test are run ahead of time. With this, the module config needs to explicitly enable each test before it is run, so it is clearer on which tests are configured to run. 
DAQ should reject invalid configurations instead of assuming some configurations. That is how we approached other configs.

grafnu:
Ok -- then maybe replace the functionality with something that rejects the
config if the core tests aren't specifically enabled/disabled. There's a
set of internal "diagnostics" that those tests are meant to represent and
there's a specific failure mode of "nothing is configured so nothing is
executed but still looks like success" that I want to avoid. Or maybe
rejects the config if *no* tests (doesn't specifically need to be the
"core" ones) are explicitly enabled/disabled.  I.e., if the enabled block
is empty that should be rejected.

henry54809:
Updated. PTAL

grafnu (approved):

PR #817, opened 2021-04-02
Read test parameters form config

anurag6:


subset/dot1x/authenticator/authenticator.py
```diff
@@ -72,21 +73,43 @@ def received_radius_response(self, payload, radius_state, packet_type):
...
     """Authenticator to manage Authentication flow"""
 
-    def __init__(self):
+    def __init__(self, config_file):
         self.state_machines = {}
         self.results = {}
         self.eap_module = None
         self.radius_module = None
         self.logger = get_logger('Authenticator')
+        self.config_file = config_file
```
grafnu, line 82: self._config_file?  (private var)

subset/dot1x/authenticator/authenticator.py
```diff
@@ -72,21 +73,43 @@ def received_radius_response(self, payload, radius_state, packet_type):
...
         self._threads = []
+        self.radius_socket_info = None
```
grafnu, line 84: private vars unless need to be exposed

anurag6: Changed. PTAL.

subset/dot1x/dot1x_config.yaml
```diff
@@ -0,0 +1,9 @@
+radius:
```
grafnu, line 1: Huh -- so my previous comment seems to have been lost (don't see it anywhere).  What' the intent of this configuration?  Is there a reason it's not using the standard module configuration mechanism?  If this is "the ability to configure the system to use an external radius server" then you shouldn't be rolling your own capability for that... (see the ipaddr module for how it's configured).

subset/dot1x/authenticator/authenticator.py
```diff
@@ -85,15 +85,19 @@ def __init__(self, config_file):
...
+    def _load_config(self):
+        self.logger.info(self._config_file)
         with open(self._config_file, 'r') as file_stream:
-            config = yaml.safe_load(file_stream)
+            full_config = json.load(file_stream)
+        config = full_config.get('modules').get('dot1x')
 
         self.logger.debug('Loaded config from %s:\n %s', self._config_file, config)
 
+        self._enabled = config['enabled']
```
grafnu, line 100: this is likely overly fragile because it will break if the key doesn't exist. Usually I'd code this as ".get('enabled')" -- unless you specifically want it to break if the field is not defined... ?

resources/setups/common/base_config.json
```diff
@@ -49,7 +49,18 @@
       "enabled": true
     },
     "dot1x": {
-      "enabled": true
+      "enabled": true,
+      "interface": "eth0",
+      "radius": {
```
grafnu, line 54: is this for the (potentially) remote radius server?  If so, maybe key with "radius_server"... ?

resources/setups/common/base_config.json
```diff
@@ -49,7 +49,18 @@
       "enabled": true
     },
     "dot1x": {
-      "enabled": true
+      "enabled": true,
+      "interface": "eth0",
```
grafnu, line 53: What's this for?  Seems wrong to be exposing this level of detail -- at what point specifically does this need to be set, and what other values might it have?

resources/setups/common/base_config.json
```diff
@@ -49,7 +49,18 @@
...
+      "radius": {
+          "secret": "SECRET",
+          "id": "02:42:ac:18:00:70",
```
grafnu, line 56: all of these except for the listen_ip seem to be the obvious defaults for running locally inside of the container?  If so, then I'd default them in the code and not here -- so this would represent explicit overrides for something outside of the conatiner.

resources/setups/common/base_config.json
```diff
@@ -49,7 +49,18 @@
...
+          "radius_socket_info": {
+              "listen_ip": "10.20.0.3",
```
grafnu, line 58: This doesn't seem like something that can be hard-coded. Semantically, what is it supposed to be?  You will likely have to find some way to dynamically determine what it should be.

grafnu (commented):
Brings up a question about what test coverage this needs (as part of the integration tests). Doesn't need to be in this PR, but the things to think about:
- skip case -- no EaPOL is detected/attempted
- pass case -- EaPOL with valid certificate exchange
- fail case -- EaPOL with invalid or error certificate exchange

But then also, for configuration, there's:
- local internal
- remote (tested as part of gateway container)

I would say that you could combine them since the goal is to just provide reasonable coverage, not complete enumeration of all possibilities, so basically down to three, so something like:

- skip
- pass w/ RADIUS in test module container
- fail w/ RADIUS in gateway module container

subset/dot1x/authenticator/authenticator.py
```diff
@@ -72,21 +73,47 @@ def received_radius_response(self, payload, radius_state, packet_type):
...
+        self._radius_socket_info = None
+        self._radius_secret = None
+        self._radius_id = None
+        self._interface = None
+        self._enabled = None
 
         self._setup()
 
+    def _load_config(self):
+        self.logger.info(self._config_file)
```
grafnu, line 93: Extra logging

subset/dot1x/authenticator/authenticator.py
henry54809: This test module won't even run when it's not enabled right? so there is no need to check for "enabled" inside the module.

resources/setups/common/base_config.json
anurag6: So, the interface is kinda important for the eap_module. I added a way to estimate the default one that's used in the tests. This can be used as an override. I will change this for dot1x test integration PR

resources/setups/common/base_config.json
anurag6: Changed.

resources/setups/common/base_config.json
anurag6: Added a way to read it by default.

resources/setups/common/base_config.json
```diff
@@ -55,7 +55,7 @@
       "enabled": true
     },
     "dot1x": {
-      "enabled": true
+      "enabled": true,
```
grafnu, line 58: extra ,

subset/dot1x/authenticator/authenticator.py
```diff
@@ -136,17 +138,46 @@ def __init__(self):
...
+        self.logger.debug('Loaded config from %s:\n %s', self._config_file, config)
+
+        self._enabled = config['enabled']
+        self._interface = config.get('interface') or get_interface_name()
+
+        radius_config = config.get('radius_server', {})
+        radius_socket_info = radius_config.get('radius_socket_info', {})
+
+        listen_ip = radius_socket_info.get('listen_ip') or get_interface_ip(self._interface)
+        listen_port = radius_socket_info.get('listen_port') or 0
```
grafnu, line 167: use default (2nd) parameter of get function (not 'or')

subset/dot1x/authenticator/authenticator.py
```diff
@@ -136,17 +138,46 @@ def __init__(self):
...
+        remote_ip = radius_socket_info.get('remote_ip') or '127.0.0.1'
+        remote_port = radius_socket_info.get('remote_port') or 1812
```
grafnu, line 169: named constant for 1812

grafnu (commented):
Still at least one unresolved comment.  Maybe go through them and make sure you've addressed them all, and if you have, add a note or explanation so it's obvious you've seen it.

Also, -- should have a test case for this (configuration) as per comment. So, for one setup it should use the internal one, and for the other it should use the gateway-based one.

daq/topology.py
```diff
@@ -424,17 +428,18 @@ def _generate_main_acls(self):
...
         for devices in self._set_devices.values():
             for device in devices:
                 if device and device.gateway:
-                    dot1x_pri_ports.extend(device.gateway.get_possible_test_ports())
-
-        self._add_acl_rule(incoming_acl, eth_type=self._DOT1X_ETH_TYPE, ports=dot1x_pri_ports)
-
-        self._add_acl_rule(secondary_acl, eth_type=self._DOT1X_ETH_TYPE,
-                           ports=list(range(1, self.sec_port)))
+                    self._add_acl_rule(incoming_acl,
```
grafnu, line 435: this nesting is getting ugly -- extract the DO1X specific stuff to a function and call that...

docker/modules/Dockerfile.networking
```diff
@@ -33,4 +33,9 @@ COPY docker/include/security/nginxfail.conf /root/nginx/
 COPY docker/include/security/nginx-site /var/www/nginx-site
 COPY docker/include/security/tlsfaux tlsfaux/
 
+COPY subset/dot1x/freeradius/users /etc/freeradius/3.0/users
```
grafnu, line 36: mmm... need to think about this. There shouldn't be a dependency from the core stuff on anything in subset/.  LIkewise, there shouldn't be a dependency on anything in the docker/ on things in subset.  What we might need to do is uplevel the common stuff to daq/misc/dot1x or something -- similar to how daq/usi is comomn between a few things -- really that should be likewise in something like daq/misc/usi (but that can be moved separately)

resources/test_site/mac_addrs/3c5ab41e8f0b/device_config.json
```diff
@@ -4,6 +4,15 @@
     "serial": "28491384"
   },
   "modules": {
+    "dot1x": {
+      "enabled": true,
+      "radius_server": {
+          "radius_socket_info": {
+              "remote_ip": "10.20.0.3",
```
grafnu, line 11: this might be unstable, but if it works it's fine.  There's no guarantee it will be at that address. Ideally DNS would work but I'm not sure if that's setup properly!

testing/test_aux.out
```diff
@@ -199,7 +202,13 @@ port-02 module_config modules
     "enabled": false
   },
   "dot1x": {
-    "enabled": true
+    "enabled": true,
```
grafnu, line 205: why is there here and in the other place?

anurag6: test_aux prints and compares module configs. I was considering dynamically editing the device_config.json file and adding GATEWAY_IP but this would get in the way.

docker/modules/Dockerfile.networking
anurag6: Pushed freeradius to misc/

daq/topology.py
anurag6: Moved to a separate method

daq/topology.py
```diff
@@ -417,24 +421,27 @@ def _add_dhcp_reflectors(self, acl_list):
...
 
+    def _add_dot1x_incoming_rule(self, incoming_acl, secondary_acl):
+        for devices in self._set_devices.values():
+            for device in devices:
+                if device and device.gateway:
+                    self._add_acl_rule(incoming_acl,
+                                       vlan_vid=self._get_port_vlan(device.port.port_no),
+                                       eth_type=self._DOT1X_ETH_TYPE,
+                                       ports=device.gateway.get_possible_test_ports())
+                    self._add_acl_rule(secondary_acl,
```
grafnu, line 432: The key to factoring out a function or bit of code is to identify repetition and reduce it as much as possible. I see the same snippet of code 4x times right now, with only minor variances -- I think it would be a fair bit cleaner to consolidate those into one function that's called 4x times.  Something like:

self._add_dot1x_rule(incoming_acl, [device.port.port_no], vlan_vid=port_vlan)

Conceptually, the consolidates the concept "the rule that handles dot1x eth" into one place... and makes it much more clear/obvious that any change needs to hit all four places (otherwise somebody needs to remember to search the code for any other uses of DOT1X_ETH_TYPE, etc...

anurag6: Refactored _add_acl_rule call into a separate method

daq/topology.py
```diff
@@ -417,24 +421,33 @@ def _add_dhcp_reflectors(self, acl_list):
...
+                if device and device.gateway:
+                    vlan = self._get_port_vlan(device.port.port_no)
+                    test_ports = device.gateway.get_possible_test_ports()
+                    if test_ports:
+                        self._add_dot1x_reflector(incoming_acl, vlan, test_ports, True)
+                    device_port = device.port.port_no
+                    if device_port:
+                        self._add_dot1x_reflector(secondary_acl, vlan, [device_port], True)
+
+    def _add_dot1x_reflector(self, acl, vlan, ports, is_trunk):
```
grafnu, line 436: generally best to avoid booleans and use named arguments, since it's more descriptive and future-proof, e.g:

`def _add_dot1x_reflector(self, acl, ports, vlan_vid=None, out_vlan=None)`

which can then translate to call like:

`self._add_dot1x_reflector(portset_acls[port_set], [self.PRI_TRUNK_PORT], vlan_vid=vlan_vid)`

It's also not a _reflector_ (quite different than what DHCP is doing), it's a allow rule, or shortcut, or something along those lines.

daq/topology.py
```diff
@@ -428,17 +428,17 @@ def _add_dot1x_incoming_rule(self, incoming_acl, secondary_acl):
...
 
-    def _add_dot1x_reflector(self, acl, vlan, ports, is_trunk):
+    def _add_dot1x_allow_rule(self, acl, ports, vlan_vid=None, out_vlan=None):
         """Add dot1x reflection rule to acl"""
-        if is_trunk:
-            self._add_acl_rule(acl, vlan_vid=vlan, eth_type=self._DOT1X_ETH_TYPE, ports=ports)
-        else:
-            self._add_acl_rule(acl, eth_type=self._DOT1X_ETH_TYPE, ports=ports, out_vlan=vlan)
+        if vlan_vid:
+            self._add_acl_rule(acl, vlan_vid=vlan_vid, eth_type=self._DOT1X_ETH_TYPE, ports=ports)
```
grafnu, line 439: keep order of arguments consistent as much as possible with the line 2 below this.   acl, ports, eth_type, *others*

grafnu (approved):

PR #818, opened 2021-04-03
Timeouts and retry support for authenticator

anurag6:


henry54809:
Maybe instead of pulling in forch’s heartbeat scheduler, you can give the “interval observable” a try. Pretty low impact.

anurag6:
> Maybe instead of pulling in forch’s heartbeat scheduler, you can give the “interval observable” a try. Pretty low impact.

Let me take a look at it. I still need to add retry support. Will check it out once that's done.

grafnu:
It's not an inherently bad idea, but I'm guessing there's still a fair
amount to be done to get 802.1x across the finish line so I'd just be wary
of distractions...

grafnu (approved):

anurag6:
Fair enough. Probably something to look at the next time.

PR #820, opened 2021-04-06
Add Varz state collector

didovesei:
Besides the VarzStateCollecotr, I also changed the way to run DAQ in MUD test, so that DAQ will not exit until the device traffic test and acl count test is done...

Also altered the sequence of starting Faucet/Gauge and generating Faucet/Gauge Config to avoid the Gauge exception which is caused by nonexistent gauge.yaml..

Btw, the codecov annotations can be turned off during code review: https://docs.codecov.io/docs/github-checks#hiding-annotations-in-the-files-view


daq/varz_state_collector.py
```diff
@@ -0,0 +1,132 @@
+"""Collect Varz states"""
+
+from __future__ import absolute_import
```
grafnu, line 3: lots of warnings here about this not being covered by tests -- do you know what's going on there?  Is this tested anywhere?

testing/test_mud.sh
```diff
@@ -47,17 +47,43 @@ function test_device_traffic {
...
+    while [ -f inst/daq.pid ]; do
+        sleep 5
+    done
+}
+
 function test_mud {
     type=$1
     echo %%%%%%%%%%%%%%%%% test mud profile $type
-    cmd/run -s device_specs=resources/device_specs/bacnet_$type.json
+    cmd/run -k -s device_specs=resources/device_specs/bacnet_$type.json & sleep 120
```
grafnu, line 74: move sleep to separate line

testing/test_mud.sh
```diff
@@ -47,17 +47,43 @@ function test_device_traffic {
...
+    python3 daq/varz_state_collector.py -y flow_packet_count_port_acl -o $metric_output_file -l $rule_filter
+    packet_count=$(jq '.gauge_metrics.flow_packet_count_port_acl.samples[0].value' $metric_output_file)
+    echo device-$device_num $type $((packet_count > 2)) | tee -a $TEST_RESULTS
+}
+
+function terminate_daq {
+    daq_pid=$(<inst/daq.pid)
+    sudo kill -SIGINT $daq_pid
+    while [ -f inst/daq.pid ]; do
+        sleep 5
```
grafnu, line 67: add some logging message here in the loop so if it gets stuck for some reason it's more obvious in the logs...

testing/test_mud.sh
grafnu: also, would this be better served by network mode (-n) rather than single-shot and hold?  Do you actually need to run the tests, or just have the system setup the network?  If you're not sure what network mode is then maybe spend just a bit of time exploring what it is and then see if it applies.
Also, you can use the "timeout" command to make some of this easier to read, maybe.  E.g., "timeout 120s XXX" -- you'll still need your kill-daq script (because it doesn't kill subshells), but it doesn't involve directly running the background process.

didovesei: So the problem I had with "-n" mode is that the devices cannot get IP addresses. I guess that's because the DHCP test is not run... As a result the devices could not generate Bacnet traffic...

daq/varz_state_collector.py
didovesei: I think it is because there is no unit test that covers this file. The reason I didn't add unit test because I think an Http server needs to be set up... Not sure if it's worth doing it as the varz collector is already being tested in test_mud.sh...

Btw, the warnings can be hidden in this page: https://docs.codecov.io/docs/github-checks#hiding-annotations-in-the-files-view

grafnu:
Ok -- sounds like a problem with network mode.  Can you file a bug against
DAQ for that so it can be fixed -- but no need to change your PR.

grafnu:
Is the file tested anywhere?  It's not just unit tests -- e.g. look in
bin/test_daq for "codecov"... should apply to all the integration tests as
well...

Ignoring warnings just makes the warnings go away, but ignores the
underlying problem so generally not the right solution!

daq/varz_state_collector.py
didovesei: Updated test_mud.sh so that it runs varz_state_collector.py with coverage instead of python. Most of the lines in varz_state_collector.py are covered by test now.

testing/test_mud.sh
didovesei: Done. PTAL..

testing/test_preamble.sh
```diff
@@ -133,3 +133,27 @@ function build_if_not_release {
...
+        echo Running Python with codecov analysis...
+        PYTHON_CMD="coverage run -a"
+    else
+        PYTHON_CMD="python3"
+    fi
+
+    export PYTHON_CMD
+}
+
+function deactivate_venv {
```
grafnu, line 153: this is odd... why do you need to do this?  Either the system should be running inside of venv or not (raw install)... why do you need to flip in and out?

didovesei: > Either the system should be running inside of venv or not
Right, makes sense. I thought only the varz_state_collector should be run in venv... Removed this function. PTAL..

grafnu (approved):

PR #832, opened 2021-04-07
Create shell of testing service (no functionality)

grafnu:


daq/session_server.py
```diff
@@ -0,0 +1,104 @@
...
+        """Start the server"""
+        LOGGER.info('Starting sesson server on ' + self._address)
+        self._server.start()
+
+    def stop(self):
+        """Stop the server"""
+        self._server.stop(grace=None)
+
+
+class SessionServerClient:
```
henry54809, line 66: Is the client here only for testing? Same as the main here?

grafnu:
Yes -- let me add a comment as such.

didovesei (approved):

PR #841, opened 2021-04-12
Updated configs for future DTS work.

grafnu:


cmd/exrun
```diff
@@ -63,7 +62,7 @@ if [ "$1" == "-h" ]; then
     skip_autostart=y
 fi
 
-export PYTHONPATH=$FORCH:$FAUCET:$MININET:$LIBS
+export PYTHONPATH=$FORCH:$FAUCET:$MININET:$ROOT
```
henry54809, line 65: Is this ROOT necessary?

cmd/start
```diff
@@ -42,5 +42,5 @@ echo Prepending $ROOT/binhack to PATH
 export OVSVSCTL_ORIG=`which ovs-vsctl`
 export PATH=$ROOT/binhack:$PATH
 
-echo Executing: $runcmd daq/daq.py $@
-$runcmd daq/daq.py $@ 2>&1
+echo Executing: $runcmd daq/entry.py $@
```
henry54809, line 45: daq.py in setup.py also needs to be renamed.

grafnu: done.

cmd/exrun
grafnu: Yes -- took me a while to figure it all out -- but the grpc generator uses absolute package names, so it's daq.proto.XXX. So, the $ROOT is necessary to find the import by absolute package name, since the basic "python daq/entry.py" runs in the subdirectory as the (nameless package root) and so doesn't no how to find an import daq.proto.XXX (it would require a relative import).  It's all just a rats-nest of dependencies!

grafnu:
Updated PTAL.

henry54809 (approved):

PR #846, opened 2021-04-14
Changes for working (not enabled) session server.

grafnu:


daq/session_server.py
```diff
@@ -20,34 +23,44 @@
 DEFAULT_SERVER_PORT = 50051
 DEFAULT_BIND_ADDRESS = '0.0.0.0'
 DEFAULT_SERVER_ADDRESS = '127.0.0.1'
-DEFAULT_RPC_TIMEOUT_SEC = 10
+DEFAULT_RPC_TIMEOUT_SEC = 600
+
+
+# pylint: disable=no-member
```
didovesei, line 29: Seems a cleaner way is to add the proto class in .pylintrcfile..

daq/session_server.py
```diff
@@ -56,12 +69,47 @@ def __init__(self, on_session=None, server_address=None, server_port=None, max_w
...
+        device_mac = request.device_mac
+        while True:
+            item = self._return_queues[device_mac].get()
+            if item is False:
+                break
+            yield item
+        LOGGER.info('Session ended for %s', device_mac)
+        del self._return_queues[device_mac]
+
     def stop(self):
```
didovesei, line 113: Should we call close_stream when the session server is stopped?

didovesei, line 113: Should we call close_stream when the session server is stopped?

didovesei: After reading the unit test, it seems that we rely on the module which has set up the server to close the stream. Closing this comment..

didovesei (approved):

daq/session_server.py
grafnu: You are correct, thanks, I forgot about that!

daq/session_server.py
```diff
@@ -20,34 +23,43 @@
 DEFAULT_SERVER_PORT = 50051
 DEFAULT_BIND_ADDRESS = '0.0.0.0'
 DEFAULT_SERVER_ADDRESS = '127.0.0.1'
-DEFAULT_RPC_TIMEOUT_SEC = 10
+DEFAULT_RPC_TIMEOUT_SEC = 600
```
henry54809, line 26: why is it 600? it seems to be a long time to wait for 1 request.

daq/session_server.py
```diff
@@ -56,14 +68,50 @@ def __init__(self, on_session=None, server_address=None, server_port=None, max_w
...
+
+    def _session_stream(self, request):
+        device_mac = request.device_mac
+        while True:
+            item = self._return_queues[device_mac].get()
+            if item is False:
+                break
+            yield item
+        LOGGER.info('Session ended for %s', device_mac)
+        del self._return_queues[device_mac]
```
henry54809, line 110: If multiple clients are connected, this line may be executed more than once right?

daq/session_server.py
grafnu: That value wasn't really used -- it was only for the testing client. I just took it out completely.

daq/session_server.py
grafnu: It's executed once per client (unique mac), yes.

henry54809: But it's possible to be executed twice under abnormal / error conditions. I'd recommend do dict.pop(mac, None) to avoid a KeyError.

grafnu:
Specifically, do you mean something like two sessions with the same MAC
address come in for whatever reason, and so there's two entries?  If so,
then I see what you're saying and you're right something should guard
against that.  The other solution would be to prevent there being two
entries in this from the get go  (e.g. it's an error to connect the second
stream if the first is still active)... need to ponder the implications of
that...

grafnu:
I checked, and there's already a check for this.  It won't let you create a
second session if one is already active (there's an assertion for it).
 The current failure mode potentially might be that in an exceptional case
the key lingers in the map and so any new session would fail... but I'm not
sure how that would happen, I will need to check on that.  So, not
something a pop() would fix, but rather maybe a try/except block.

PR #847, opened 2021-04-14
Adjust codecov

grafnu:


codecov.yml
```diff
@@ -3,6 +3,8 @@ coverage:
     project:
       default:
         target: 80%
+        threshold: .5%
     patch:
       default:
+        target: 80%
         threshold: .5%
```
henry54809, line 10: I wonder if these .5% are accepted as valid thresholds

grafnu: Do you mean that it needs to be an integral number, or something else?  Something isn't right and I'm just playing around to see what it might be. I'll try 1% to see what it does...

grafnu: Yup -- that was it. I changed it to 1% and it passed.

PR #853, opened 2021-04-29
Disable TLS test

anurag6:


config/modules/all.conf
```diff
@@ -4,7 +4,6 @@ include ${DAQ_LIB}/config/modules/host.conf
 # All contributed modules.
 include ${DAQ_LIB}/subset/switches/build.conf
 include ${DAQ_LIB}/subset/bacnet/build.conf
-include ${DAQ_LIB}/subset/security/build.conf
```
grafnu, line 7: comment out with TODO indicating roughly what it's all about.

grafnu (commented):
I don't think this is going to be sufficient... I'm pretty sure it'll need some adjustments elsewhere to get the tests to pass...

PR #855, opened 2021-05-03
Check for stale/disconnected session server clients

grafnu:


didovesei (approved):

daq/session_server.py
```diff
@@ -101,18 +112,42 @@ def _init_session(self, request):
     def _session_stream(self, request):
         device_mac = request.device_mac
         while True:
+            self._last_streamed[device_mac] = time.time()
```
henry54809, line 115: I think this line should go after the next line like this
<pre>
item = self._return_queues[device_mac].get()
self._last_streamed[device_mac] = time.time()
</pre>
since queue.get is a blocking call

grafnu: Fixed.

PR #856, opened 2021-05-05
Add DAQ_LOG env var for file logging. Move DAQ run cleanups to python

henry54809:


daq/device_report_client.py
```diff
@@ -52,7 +52,7 @@ def cancel_request():
         threading.Thread(target=self._port_event_handler,
                          args=(callback, result_generator, cancel_request)).start()
 
-    def terminate(self):
+    def stop(self):
```
henry54809, line 55: This change is needed to match the new session server.

testing/test_aux.sh
```diff
@@ -261,7 +261,7 @@ monitor_log "Port 1 dpid 2 is now active" "sudo ifconfig faux down;sleep 15; sud
 monitor_log "Target device 9a02571e8f00 test hold running" "sudo ifconfig faux down"
 rm -r inst/run-*
 cmd/run -s -k
-disconnections=$(cat inst/cmdrun.log | grep "Port 1 dpid 2 is now inactive" | wc -l)
+disconnections=$(cat inst/daq.log | grep "Port 1 dpid 2 is now inactive" | wc -l)
```
grafnu, line 264: I would expect this change (cmdrun.log -> daq.log) to have ramifications elsewhere... So, nothing to change in this PR, but just for us to be aware of that this change might cause cascading failures.  Once this PR is submitted, can you do an update on forch to point at the new version (of DAQ) and make sure it still works?  Basically, I think there are some implicit dependencies buried somewhere in the forch integration tests...

grafnu (approved):

PR #859, opened 2021-05-07
Updates for dynamix vxlan for DTS

grafnu:


config/modules/dts.conf
```diff
@@ -0,0 +1,12 @@
...
+# Tests to build.
+build ${DAQ_LIB}/docker/modules
+
+# Default built-in tests.
+add pass first
+add fail first
+add ping first
+
+# Additional base modules
+include ${DAQ_LIB}/subset/pentests/build.conf
```
henry54809, line 12: Doesn't this also include the discover test which is in a docker container?

henry54809: I've been using 
<pre>
add pass
add fail
add ping
add nmap
</pre>

dts/start_service
```diff
@@ -0,0 +1,74 @@
...
+sudo /usr/share/openvswitch/scripts/ovs-ctl start
+
+echo Configuring for vxlan endpoint ip $ip_address
+echo switch_setup.endpoint.ip=$ip_address >> $DYNAMIC_CONFIG
+
+echo Starting daq, output in $DAQ_ROOT/inst/cmdrun.log
+PYTHONPATH=$PYTHONPATH \
+    PATH=$DAQ_ROOT/mininet:$PATH \
+    FAUCET_EVENT_SOCK=$DAQ_EVENT_SOCK \
+    $DAQ_ROOT/cmd/start $BASE_CONFIG $DYNAMIC_CONFIG -k $no_test > $DAQ_ROOT/inst/cmdrun.log 2>&1 &
```
henry54809, line 72: is the -k necessary?

proto/session_server.proto
```diff
@@ -51,4 +68,5 @@ message SessionResult {
  */
 service SessionServer {
   rpc StartSession(SessionParams) returns (stream SessionProgress) {}
+  rpc ConfigureEndpoint(TunnelEndpoint) returns (SessionResult) {}
```
henry54809, line 71: Is this extra? it's also not defined anywhere.

.gitignore
```diff
@@ -5,6 +5,8 @@
 .build_built
 .project
 
+/tmp/
```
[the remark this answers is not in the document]
grafnu: Just playground use... sometimes I need to stash files somewhere, and this makes it doable without confusing git.

daq/session_server.py
```diff
@@ -54,16 +53,17 @@ def StartSession(self, request, context):
...
+    def __init__(self, on_session=None, server_address=None, server_port=None, local_ip=None):
         self._return_queues = {}
         self._last_streamed = {}
         self._done_callbacks = {}
         self._lock = threading.Lock()
         self._on_session = on_session
-        self._server = grpc.server(
-            futures.ThreadPoolExecutor(max_workers=max_workers or DEFAULT_MAX_WORKERS))
+        self._local_ip = local_ip
+        self._server = grpc.server(futures.ThreadPoolExecutor())
```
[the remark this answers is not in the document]
grafnu: Unless we have some specific reason, I'd leave fixing the value out b/c we don't really know what to fix it at, so a default might actually make sense. Otherwise we're surely making an arbitrary decision for something... rather than making no comment on the subject.

config/modules/dts.conf
grafnu: fixed

proto/session_server.proto
grafnu: removed

henry54809 (approved):

dts/start_service
grafnu: apparently not. removed.

PR #861, opened 2021-05-09
Add mininet retry, fix pass-on-exception, properly handle partial remote state updates

grafnu:


daq/runner.py
```diff
@@ -1009,6 +1016,11 @@ def _calculate_device_result(self, test_results):
                 if result not in ('pass', 'skip'):
                     failed = True
 
+        if len(device.host.enabled_tests) != len(processed):
```
henry54809, line 1019: I think these 2 will never be equal since enabled_tests doesn't contain the startup stages like "acquire", "base" which are still included in module_results. A simple set include check should suffice.

daq/test_modules/external_module.py
```diff
@@ -148,13 +148,16 @@ def _finalize(self):
             self.pipe = None
         return_code = self.host.terminate()
         LOGGER.info('%s test host finalize %s', self, return_code)
+        modified_code = return_code if self.test_name != 'fail' else not return_code
```
henry54809, line 151: What's the purpose of this? It's going to create so much confusion.

grafnu: This particular change is moving this check from somewhere else to here.  The purpose of the 'fail' module is to test that a failing module is handled correctly. Yes, it's awkward now (very old legacy test), but I'm not inclined to clean it up as part of this PR. I added a TODO to replace it with a proper unit test for a failing module.

daq/runner.py
grafnu: But it does work... ?  It's not listing all tests that are included in module_results, only the ones that have a result_code, so maybe that's why it works?

grafnu:
Updated PTAL

Issue #845, opened 2021-04-13
Include the possibility to setup an NTP server using a hostname or IP address rather than have a hard coded NTP provided by DHCP


pisuke:

grafnu:
NB: This would be very similar functionality to what has been recently been done for 802.1x, which either uses the default internal radius server, or can be configured to use an external radius server (as configured by IP address somewhere).

noursaidi:
I think this is referring to device configuration of which NTP server to use rather than a test module. We don't strictly need a local DAQ NTP server, as the tests only check that (i) protocol is NTPv4 and (ii) the time between the NTP client and server is small enough that the client is considered to be synchronised. The second test currently also checks that the device is only communicating with the local DAQ NTP server as a pass condition, which is the functionality up for debate.

The local DAQ NTP server runs on a random IP address with each run, so cannot be configured with this. I think the best solution is assigning a hostname to the NTP server (e.g daqntp.local?) in Dnsmasq.  Then update the documentation so that the test operator configures the device to use this hostname as the NTP server.  

The benefit is that this also demonstrates that the NTP server on the device can be configured which is a desired behaviour. Alternatively, to keep tests testing one thing only, could become an INFO test and/or the NTP tests could allow communication with any NTP servers. However the the test logic will need to be revamped to track communication with different servers which may be in an NTP pool.

pbatta:
The idea of pointing the device-under-test to the DAQ NTP server by setting
the hostname does sound good. So the existing tests would remain, and users
would (based on the capabilities of the device) either set the NTP server,
or let the device learn from DHCP based on what the device is capable of,
correct?

noursaidi:
My apologies @pbatta completely missed the notification. 

I suggest this test uses the hostname only. I understand the expectation is now that NTP through DHCP is not required? If we leave the DHCP support in there, I suggest we add add a test item which identifies if the device was configured using DHCP or via hostname. This would be a simple test, as it could just look for a DNS request for the hostname

grafnu:
The problem is that the word "required" is not as binary as you might
think.  The NTP-through-DHCP is an "intermediate" step that is OK for now,
but we should leave the other test in there.  Ideally, they would both be
checked and have two separate line items... so the device is tested and
says if it passed "NTP/DNS" or "NTP/DHCP".  Basically, the complete
requirement is "do NTP-through-DHCP unless explicitly acknowledged that it
only does NTP-through-DNS" -- so there would be a configuration parameter
[...]

PR #862, opened 2021-05-10
Update NTP tests to use hostname 

noursaidi:
Additional discussion on issue #845 

Updates the NTP tests to use the hostname rather than DHCP by adding a hostname `ntp.daqlocal` and the NTP server IP address from DHCP

### DIFF
--- a/daq/report.py
+++ b/daq/report.py
@@ -270,7 +270,8 @@ class ReportGenerator:
 
                 # The device overall fails if any result is unexpected
                 if result_dict["result"] != required_result:
-                    passes = False
+                    if required_result == 'notfail' and result_dict["result"] == 'fail':
+                        passes = False
 
                 if result_dict["result"] == 'gone':
                     gone = True
@@ -308,7 +309,7 @@ class ReportGenerator:
             total = 0
 
             results = [[0, 0, 0] for _ in range(len(self._expected_headers))]
-            result = self._NO_REQUIRED  # Overall category result
+            result = self._NO_REQUIRED  # Overall category result is n/a if no tests
 
             for test_name, result_dict in self._results.items():
                 test_info = self._get_test_info(test_name)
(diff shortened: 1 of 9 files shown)


docker/include/bin/start_faux
```diff
@@ -156,11 +156,10 @@ elif [ -n "${options[bacnetfail]}" ]; then
          FauxDeviceEngine.EntryPoint $local_ip $broadcast_ip "Faux-Device-Fail.json"  &
 fi
 
-# NTPv4 query to the NTP server learnt from DHCP.
+# NTPv4 query to the local NTP server
 # NTPv3 query to the IP of time.google.com (since resolv.conf is modified by other tests)
 if [ -n "${options[ntpv4]}" ]; then
-    dhcp_ntp=$(fgrep NTPSERVERS= /run/ntpdate.dhcp)
-    ntp_server=`echo $dhcp_ntp | cut -d "'" -f 2`
+    ntp_server=ntp.daqlocal
```
pbatta, line 162: since ntp.daqlocal is used a couple of times, make it a named variable in this file? (LOCAL_NTP_SERVER_NAME=ntp.daqlocal or something like that)

docker/include/network/scripts/start_networking
```diff
@@ -31,10 +31,12 @@ subnet=${IP_SUBNET:-$random_subnet}
 echo Configuring with subnet 10.20.$subnet.XX
 echo dhcp-range=10.20.$subnet.100,10.20.$subnet.254 >> /etc/dnsmasq.conf
 
-# Use ourself as the DNS and NTP server explicitly specified
+# Use ourself as the DNS server explicitly specified
 ip addr add 10.20.$subnet.2 dev $LOCAL_IF
 echo dhcp-option=6,10.20.$subnet.2 >> /etc/dnsmasq.conf
-echo dhcp-option=42,10.20.$subnet.2 >> /etc/dnsmasq.conf
```
pbatta, line 37: Should we delete this? Leaving this in place would allow the DUT to either learn NTP from DHCP option, or have it statically configured, so we'll cover both options?

noursaidi: Yes, on the issue  #845 there's been additional comments from Trevor, as well as your comment from last week on having both test cases, so this PR will be updated accordingly

PR #864, opened 2021-05-13
Create a "native gateway" config for DAQ, for the ATA setup (test appliance)

grafnu:


grafnu:
Sorry this PR is so long/large -- it just ended up touching a lot of little things.  The main concept that's introduced is that of a "native gateway", which is like the regular DAQ test gateway, except that it deals with a native vlan (rather than a tagged trunk), and is instantiated before actual devices are discovered (rather than on-demand when devices are discovered).

cmd/faux
```diff
@@ -119,8 +119,8 @@ pid=$(docker inspect --format="{{ .State.Pid }}" $cid)
 
 echo DAQ autoclean docker kill $CONTAINER
 
-if [ -n "$realintf" ]; then
-    echo Moving $realintf into $pid...
+if [[ -n $realintf ]]; then
+    echo Moving real $realintf into $pid...
```
anurag6, line 123: NIT: real interface?

daq/container_gateway.py
```diff
@@ -11,17 +11,16 @@
 
 LOGGER = logger.get_logger('gateway')
 
-
```
henry54809, line 14: Why is this line removed? It's standard to have blank 2 lines before class.

docker/include/network/scripts/relay_arp
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env python3
+
+from scapy.all import sendp,sniff,ARP
```
henry54809, line 3: Can we also add this file to bin/check_style.

daq/runner.py
```diff
@@ -296,6 +317,22 @@ def initialize(self):
 
         LOGGER.debug('Done with initialization')
 
+    def _create_stream_monitor(self):
+        return stream_monitor.StreamMonitor(idle_handler=self._handle_system_idle,
+                                            loop_hook=self._loop_hook,
+                                            timeout_sec=20)  # Polling rate
+
+    def _create_native_gateway(self):
+        if not self._native_vlan:
```
henry54809, line 326: This function is odd; this will probably be even more convoluted to anyone not familiar with DAQ.  Is there another path that can be used to start a native gateway?

grafnu: Yes -- I agree that was particularly messy. I've refactored everything, so hopefully it's better now.

daq/runner.py
```diff
@@ -32,6 +32,11 @@
 
 LOGGER = logger.get_logger('runner')
 
+# Totally not valid MAC used for key of a fake device.
+FAKE_DEVICE_MAC = '00:00:00:00:00:00'
```
henry54809, line 36: This is not needed now right

henry54809 (approved):

daq/test_modules/ipaddr_module.py
```diff
@@ -38,14 +37,8 @@ def __init__(self, host, tmpdir, test_name, module_config):
...
         ]
-        self._logger = logger.get_logger('ipaddr_%s' % self.host_name)
-        log_folder = os.path.join(self.tmpdir, 'nodes', self.host_name, 'tmp')
-        os.makedirs(log_folder)
-        log_path = os.path.join(log_folder, 'activate.log')
-        self._file_handler = logging.FileHandler(log_path)
-        formatter = logging.Formatter(_LOG_FORMAT)
-        self._file_handler.setFormatter(formatter)
-        self._logger.addHandler(self._file_handler)
+        log_file = os.path.join(self.tmpdir, 'nodes', self.host_name, 'tmp', 'activate.log')
```
henry54809, line 40: Why is this changed? There’s no need to have a separate custom logging handler

grafnu: I separated out the logging changes to #865 -- we can address/finalize that first, and then come back to this PR.

PR #865, opened 2021-05-17
Consolidate module logging, and some logging cleanup

grafnu:


daq/logger.py
```diff
@@ -44,14 +55,18 @@ def get_logger(name=None):
...
     return LOGGERS[name]
 
 
-def set_config(level='info', fmt=None, datefmt=None):
+def set_config(level='info'):
     """Sets config for all loggers"""
     root_logger = logging.getLogger()
     root_logger.setLevel(level)
-    formatter = logging.Formatter(fmt=fmt, datefmt=datefmt)
+    formatter = logging.Formatter(fmt=_LOG_FORMAT, datefmt=_DATE_FORMAT)
```
henry54809, line 69: This `formatter` can be replaced with _FORMATTER

daq/test_modules/base_module.py
```diff
@@ -30,6 +32,11 @@ def __init__(self, host, tmpdir, test_name, module_config):
         self.params = None
         self.start_time = None
 
+    def get_logger(self, log_name):
```
henry54809, line 35: This function can probably just stay in ipaddr_module for now. When there's a need for other in line modules, we can create a new InlineModule parent class to host this function. Storing it in base_module gives the illusion that all test modules use this extra logger to write out activate.log where in fact only the inline modules do.

daq/logger.py
henry54809: actually why is _FORMATTER not using the date format?

grafnu: consolidated

henry54809 (approved):

daq/test_modules/base_module.py
grafnu: The goal was to get it out of ipaddr since I know it's not ipaddr-specific, and it does have more general utility. There are potentially reasons why a regular module might want to use this "to a host specific file" facility, not just for inline modules I'd wager. To avoid confusion with activate.log (which was confusing to me as well even just for ipaddr), I renamed it to module.log. (Yes, I know that functionally for ipaddr this is somewhat the same as activate.log -- but I think it's ultimately more confusing to have activate.log in two places.)

PR #854, opened 2021-04-30
Disable tls test

anurag6:

PR #869, opened 2021-06-06
Fix TLS test bug and re-enable in CI

noursaidi:
This PR reverts PR #854 and resolves the underlying issue.

The problem was JDK11.0.11 update disabled TLS1 by default (https://www.oracle.com/java/technologies/javase/11-0-11-relnotes.html) This PR includes a script `subset/security/enable_tls1.sh` to update  the configuration at container build. The TLS test container uses Ubuntu 20.04 and a newer version of OpenSSL which also default disables older/insecure protocols, so TLSv1 has also been enabled in OpenSSL.

This bug meant the test always skipped, regardless of whether the DUT was running TLSv1 and the result should have been a pass/fail. The faux device currently run an older version, so is okay for now.


grafnu (approved):

grafnu:
The MUD test failure has been fixed in another PR, so might try rebasing this and see if it does better!

noursaidi:
Thanks, recommitted and passes the tests now

PR #876, opened 2021-06-10
Change settle_sec to 0

grafnu:
One thing necessary for testing at scale.  Doesn't seem to break anything.

### DIFF
--- a/config/system/default.yaml
+++ b/config/system/default.yaml
@@ -13,7 +13,7 @@ switch_setup:
   uplink_port: 7
 
 # Time to let the network settle after some operations.
-settle_sec: 5
+settle_sec: 0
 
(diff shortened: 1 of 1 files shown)


config/system/default.yaml
```diff
@@ -12,9 +12,6 @@ switch_setup:
   of_dpid: 2
   uplink_port: 7
 
-# Time to let the network settle after some operations.
-settle_sec: 5
```
pbatta, line 16: Also remove other references? (proto/system_config.proto, daq/network.py etc?)

grafnu:
Not now -- this is just the default configuration but we might want to turn
it on for some specific install.

anurag6 (approved):

PR #883, opened 2021-06-16
Find static ip devices if external dhcp wasn't detected. 

henry54809:
Find static ip devices using arp-scan if external dhcp wasn't detected / timedout.

### DIFF
--- a/daq/base_gateway.py
+++ b/daq/base_gateway.py
@@ -231,24 +231,8 @@ class BaseGateway(ABC):
 
     def _get_scan_interface(self):
         return self.host, self.host_intf
-
-    def discover_host(self, subnets: List[ip_network], mac: str, callback: Callable, vid=None) -> ip_address:
-        cmd = 'arp-scan --retry=2 --bandwidth=512K --interface=%s --destaddr=%s %s %s'
-        host, intf = self._get_scan_interface()
-        LOGGER.info('Starting host discovery for %s', mac)
-        def hangup_callback(log_fd, log_file):
-            log_fd.close()
-            with open(log_file, 'r') as fd:
-                lines = fd.read().split('\n')
-                for line in lines:
-                    device_ip = process_line(line)
-                    if device_ip:
-                        LOGGER.info('Host discovery for %s completed. Found ip %s.',
-                                    mac, device_ip)
-                        return callback(device_ip)
-            LOGGER.info('Host discovery for %s completed. Found no ip.', mac)
-            callback(None)
-
+    
+    def _discover_host_hangup_callback(self, log_fd, log_file, callback):
         def process_line(line):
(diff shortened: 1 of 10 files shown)


daq/base_gateway.py
```diff
@@ -224,6 +229,49 @@ def terminate(self):
                 LOGGER.error('Gateway %s terminating dummy: %s', self.name, e)
                 LOGGER.exception(e)
 
+    def _get_scan_interface(self):
+        return self.host, self.host_intf
+
+    def discover_host(self, subnets: List[ip_network], mac: str, callback: Callable, vid=None) -> ip_address:
```
grafnu, line 235: is this discover_host or discover_hosts (plural)?  Also, does the -> ip_address at the end imply that it's returning an ip_address... ?  Does it actually return anything (seems like it's callback-oriented, not return-oriented...).

daq/base_gateway.py
```diff
@@ -224,6 +229,49 @@ def terminate(self):
...
+        cmd = 'arp-scan --retry=2 --bandwidth=512K --interface=%s --destaddr=%s %s %s'
+        host, intf = self._get_scan_interface()
+        LOGGER.info('Starting host discovery for %s', mac)
+        def hangup_callback(log_fd, log_file):
```
grafnu, line 239: Do these functions (hangup_callback, process_line) need to be inlined with the main function (discover_host)?  I know it makes some things easier, but typically I find it harder to read (overall function is ~40 lines when it could be broken up into smaller subunits). Seems like hangup_callback doesn't have much borrowed from the discover_host context (just callback, which could be passed with a lambda).

henry54809: I can clean these up but is there any other concern with this way of discovering hosts? It seems to work very well (for ipv4 at least).

grafnu: Ok -- just making sure I understand what's going on.  The basics are that when a host times out waiting for host DHCP it will call this function with the mac address of the host, and then this calls arp-scan on that host which will send it an ARP to try and figure out it's IP address?   It's not, for example, trying to periodically scan all IP addresses in the subnet or something like that?

If so (hitting a single host), then I think it's perfectly fine.

daq/base_gateway.py
grafnu: If this is in fact discover_host (as I think it is after I looked at it), then the mac argument should likely be before the subnets argument (which is what caused some of my confusion).

daq/base_gateway.py
henry54809: So yes, when a host times out when using external DHCP, it'll try to figure out host's IP using arp-scan. Other dhcp times out aren't affected.

henry54809:
PTAL

grafnu (approved):

PR #884, opened 2021-06-18
Tweaks for DTS testing

grafnu:


daq/base_gateway.py
```diff
@@ -258,7 +258,7 @@ def process_line(line):
 
     def discover_host(self, mac: str, subnets: List[ip_network], callback: Callable, vid=None):
```
henry54809, line 259: The vid arg is still here even though it’s not being used.

grafnu:
Greatly simplified to just be the changes required for DTS.

henry54809 (approved):

PR #887, opened 2021-07-01
Additional k8s files

henry54809:


dts/dts.yaml
```diff
@@ -0,0 +1,3 @@
+include: ${DAQ_LIB}/config/system/dts-base.yaml
```
grafnu, line 1: why split out dts-base.yaml?  Is it used somewhere else? (Or intended to be used somewhere else)  I'm wondering some then too if the specification should be inverted, e.g. "dts.yaml" is the base, and then something like the kustomization.yaml file would have a derivative (dts-cluster.yaml) or something.  Having both a "dts.yaml" and "dts-base.yaml" seems odd to me for some reason (without knowing specifically why it might be like that)...

henry54809: OK. I can leave the original dts.yaml as is and rename the kustomization's dts yaml to dts-cluster.yaml. I intended the 2 to be be named different so it's clearer that one's built into the container and the other can be used to override at deploy time.

grafnu: sgtm -- I think I just wasn't sure how to differentiate "dts" from "dts-base" -- so the other naming is a bit more clear!

grafnu (approved):

PR #893, opened 2021-07-14
Basic report framework

grafnu:


daq/runner.py
```diff
@@ -1075,6 +1078,13 @@ def _send_device_result(self, device, test_results):
         except Exception as e:
             LOGGER.error("Failed to send device results for device %s: %s ", device, e)
 
+    def report_sink(self, report_dict):
```
henry54809, line 1081: If it's not consumed by some grpc client, does it really have to be in the protobuf? The existing dict / a dataclass to replace https://github.com/faucetsdn/daq/blob/master/daq/report.py#L240 should suffice.

grafnu:
It will be consumed by some client, even if it's not gRPC (protobuf isn't
only useful for grpc). I want something that is well-structured and
cross-language, so something more than an arbitrary dict (unstructured),
and not just a dataclass (python-only).

grafnu:
Any additional comments?

proto/report.proto
```diff
@@ -0,0 +1,37 @@
...
+}
+
+message TestModule {
+  // Line items test results from this module.
+  map<string, TestItem> tests = 1;
+}
+
+message TestItem {
+  // Test result (pass/fail)
+  string result = 1;
```
henry54809, line 21: Should that be in enum? But we can revisit this later.

henry54809 (approved):

PR #894, opened 2021-07-14
Remove faucet version workaround

grafnu:


henry54809:
There still should be a networkx==2.5.1 version pin; otherwise a future release of networkx with numpy requirements will break the build again

PR #895, opened 2021-07-16
Pop VLANs while pushing dot1x packets out of device port

anurag6:
Remove VLAN tags on outgoing EAP packets on device port.

### DIFF
--- a/daq/topology.py
+++ b/daq/topology.py
@@ -442,14 +442,16 @@ class FaucetTopology:
                         self._add_dot1x_allow_rule(incoming_acl, test_ports, vlan_vid=vlan)
                     device_port = device.port.port_no
                     if device_port:
-                        self._add_dot1x_allow_rule(secondary_acl, [device_port], vlan_vid=vlan)
+                        self._add_dot1x_allow_rule(secondary_acl, [device_port], in_vlan=vlan)
 
-    def _add_dot1x_allow_rule(self, acl, ports, vlan_vid=None, out_vid=None):
+    def _add_dot1x_allow_rule(self, acl, ports, vlan_vid=None, out_vid=None, in_vlan=None):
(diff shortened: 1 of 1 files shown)


daq/topology.py
```diff
@@ -439,17 +439,21 @@ def _add_dot1x_incoming_rule(self, incoming_acl, secondary_acl):
...
+                        self._add_dot1x_allow_rule(incoming_acl,
+                                                   test_ports, vlan=vlan, action='vlan_vid')
                     device_port = device.port.port_no
                     if device_port:
-                        self._add_dot1x_allow_rule(secondary_acl, [device_port], vlan_vid=vlan)
+                        self._add_dot1x_allow_rule(secondary_acl,
+                                                   [device_port], vlan=vlan, action='in_vlan')
 
-    def _add_dot1x_allow_rule(self, acl, ports, vlan_vid=None, out_vid=None):
+    def _add_dot1x_allow_rule(self, acl, ports, vlan=None, action=None):
```
grafnu, line 449: this is a somewhat clumsy way of doing this IMHO -- I'd much rather it keep the named arguments approach and just have vlan_vid=, out_vid=, in_vlan=  and use that for the if statements below.   Was there a particular reason you wanted to expand it out into the case-statement format?  Functionally the same, but it pushes more of the complexity to the function rather than the call site.

anurag6: Lint complains if I simply add a variable. I just did this to get past it but I guess I could add an ignore flag as well. Just a tradeoff at the end of the day. Should I be going the ignore flag route?

grafnu:
What do you mean by "simply add a variable"  --- what is lint complaining
about?  There's code all over the place that does something similar, so I
don't quite understand what the specific lint complaint is!

grafnu (approved):

Issue #897, opened 2021-07-19
Make cloud test initial configuration easier


noursaidi:
A few changes are required to the cloud tests module to make initial configuration easier
- relocate the site model in the site path rather than inst/config directory
- use reflector key from the `SITEMODEL/validator` directory where it currently lives rather than

PR #898, opened 2021-07-19
cloud tests documentation update

noursaidi:
A brief update to the cloud tests documentation to describe the steps required to configure the cloud tests module as it stands at present.

I've created issue #897 which describes two changes to the module to make initial configuration easier, which if/when implemented will require a small change to the documentation


config/modules/cloudonly.conf
```diff
@@ -0,0 +1,15 @@
+# Cloud only tests
+
+# Default built-in tests.
+include ${DAQ_LIB}/config/modules/host.conf
```
grafnu, line 4: If this is really intended to be cloud-only, then this shouldn't include the base host tests and remove others.  What happens if something else is added into host.conf?  Then the logic would end up not being "cloud-only"...

grafnu:
Mostly looks good, just one outstanding comment. Let's try to get this cleaned up and submitted to get it out of the queue?

config/modules/cloudonly.conf
noursaidi: I've changed this to only build core modules and cloud test module

config/modules/cloudonly.conf
```diff
@@ -0,0 +1,5 @@
+# Cloud only tests
+
+build ${DAQ_LIB}/docker/modules
+include ${DAQ_LIB}/usi/build.conf
```
grafnu, line 4: Why USI?  That's not cloud oriented... ?  Is it because there's some part of the system that always tries to run the USI test, even if not explicitly enabled?

Also, can you just name the file "cloud.conf" -- just keeping in the minimalist spirit of the other files in that directory...

grafnu (approved):
Just a few minor tweaks -- but feel free to submit after you make them (or if they changes don't make sense you can submit anyway...). Just minor suggestions.

config/modules/cloudonly.conf
noursaidi: Re USI - yes, I'd assumed with switch control being abstracted it was the module used for controlling the virtual switch so assumed it's needed. I'll test without it today and omit if not required

noursaidi:
Just tested without USI, and at the moment it needs the USI module to run, so I've left it in the cloud config. 

```
nour@ubun2:~/daq2/daq$ sudo cmd/run 
[sudo] password for nour: 
Activating venv
Flattening config from local/system.yaml into inst/config/system.conf
Starting Fri 10 Sep 19:32:25 BST 2021
Clearing previous state...
Activating venv
Flattening config from local/system.yaml into inst/config/system.conf
HEAD is now at 9a1c90fc Merge pull request #3837 from gizmoguy/ci-improvements
HEAD is now at a6c78b5f Revert faucet version
HEAD is now at b56721d FIx minor variable typo
Release version
Cleaning bridge pri...
Cleaning bridge sec...
ovsdb-server is running with pid 21883
ovs-vswitchd is running with pid 21947
No external switch model specified.
Error: No such container: daq-usi
DAQ autostart gcp_cred= cmd/usi
Starting USI in debug mode
Unable to find image 'daqf/usi:latest' locally
docker: Error response from daemon: manifest for daqf/usi:latest not found: manifest unknown: manifest unknown.
See 'docker run --help'.
Unable to find image 'daqf/usi:latest' locally
docker: Error response from daemon: manifest for daqf/usi:latest not found: manifest unknown: manifest unknown.
See 'docker run --help'.
nour@ubun2:~/daq2/daq$ 

```

grafnu:
SGTM -- just wanted to make sure it was "needed"!

PR #900, opened 2021-08-03
gateway allocation cleanup

henry54809:


daq/runner.py
```diff
@@ -192,7 +198,7 @@ def __init__(self, config):
         self._cleanup_previous_runs()
         self._init_test_list()
         self._target_set_queue = []
-        self._max_hosts = self.run_trigger.get('max_hosts') or float('inf')
+        self._max_hosts = self.run_trigger.get('max_hosts', float('inf'))
```
grafnu, line 201: was there a specific need for this change?  The difference is in the behavior when max_hosts == 0, where in the former case it would convert that to inf, while in the latter it would make it zero.  Was there a use-case where you needed to be able to configure "max_hosts" to 0?  The reason for the former logic is that it makes it possible to override some upstream setting to "inf" by setting the value to 0 (which is the default case if anything is sorted in a protobuf).

henry54809: This config is not a protobuf object so there is no default 0 value for max_hosts. I think it's best to not have any override for values that don't make sense(in your case, a negative value for max_hosts is also not override). We need to expand the validation of config in the future so daq can just throw an error if some config values don't make sense.

grafnu:
At this level in the code there's no explicit recognition that "config is
not a protobuf" -- that's a true statement in the current implementation,
but I don't find it a particularly compelling argument. I don't know what
the statement "not have any override for values that don't make sense"
means.  The question at hand is "how do you override something to 'default'
behavior" -- in the previous version, you could override to default by
specifying 0 (which is a nonsensical value for "max hosts" so is
semantically unique" -- so really that's my question (proposed mechanism
for a config file to override an inherited config).

henry54809:
PTAL. I restored that max_hosts. we can push off that discussion until we are ready to refactor configs.

grafnu (approved):

PR #901, opened 2021-08-05
Updates for ATA use cases

grafnu:


docker/include/network/scripts/start_networking
```diff
@@ -15,7 +15,7 @@ while ! ip link show $EXT_INTF; do
     sleep 5
 done
 
-if [[ $EXT_MAC ]]; then
+if [[ -n $EXT_MAC && $EXT_MAC != None ]]; then
```
henry54809, line 18: I don't see where this is being used and why is it possible for EXT_MAC be None?

daq/runner.py
```diff
@@ -622,6 +627,10 @@ def _target_set_trigger(self, device, remote_trigger=False):
             LOGGER.debug('Target device %s already queued', device)
             return
 
+        if device.mac in self._target_set_block:
```
henry54809, line 630: I think something like target_set_triggered will be better than target_set_block(it raises a new concept of device blocking which confuses me at first)

daq/runner.py
```diff
@@ -985,18 +998,15 @@ def _target_set_finalize(self, device, result_set, reason):
...
                 self._linger_exit = 1
         self._result_sets[device] = result_set
 
-        if self.run_limit and self.run_count >= self.run_limit and self.run_tests:
+        if self._run_limit and self._run_count >= self._run_limit and self._run_tests:
             LOGGER.warning('Suppressing future tests because run limit reached.')
-            self.run_tests = False
-        if self.single_shot and self.run_tests:
-            LOGGER.warning('Suppressing future tests because test done in single shot.')
-            self._handle_faucet_events()  # Process remaining queued faucet events
```
henry54809, line 998: I think this last faucet event processing can be left in.

config/system/ata.yaml
```diff
@@ -16,7 +19,7 @@ switch_setup:
 
 run_trigger:
   native_vlan: 122
-  max_hosts: 8
+  max_hosts: 3
```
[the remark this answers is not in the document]
grafnu: For ATA, it's limiting the number of tests that can run inside of one portset -- there is always only one port-set in that case (since it's a shared network). I'd like to leave it at 3 (because that's what I've tested) until we can fix the limiting factor and then bump it up.

docker/include/network/scripts/start_networking
grafnu: It gets converted from ext_mac to EXT_MAC in the code -- (in base_gateway.py _get_env_vars() does the case conversion). So, it's set in the configuration as ext_mac, and then gets passed along as EXT_MAC (like a few other parameters). It can be None because somewhere in there python is automatically converting a None value to "None". The alternate fix would be to add an if statement in the code that says "if not value then don't add the env var setting" or sets the value to "". Either way, there's an if statement somewhere, just not sure where (I was going with the principle of the most-focused change, but there are other principles to follow).  What do you think?

daq/runner.py
grafnu: The intent was that "triggered" means that it's been triggered for a test run, while "blocked" is to mean that it's blocked from future test runs. In single-shot mode, triggering causes it to be blocked, but there's cases where triggered != blocked, and also where we might want to pre-block devices so they are never triggered.  So yes, it was intentionally a new concept... not sure how to make it less confusing.

daq/runner.py
grafnu: But _should_ it be left in (and if so, why)?  I was trying to figure out what that code was for and it seemed out of place, so I was trying to clean things up a bit.

docker/include/network/scripts/start_networking
henry54809: I think we can follow https://github.com/faucetsdn/daq/blob/04967711fc58c3101c2449d09dbc9dcef95df21e/daq/test_modules/external_module.py#L110 
to convert None to "" and check here [ ! -z $EXT_MAC ]

daq/runner.py
henry54809: What's the case where blocked != triggered under single shot? The variable can be scoped to be single_shot_target_set_triggered. There isn't a way to pre-block devices either so I don't think we need to introduce this "blocked" concept at the moment.

config/system/ata.yaml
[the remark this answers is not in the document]
henry54809: Yes I understand there is only 1 portset for ata but there are 4 ports available on that portset for testing. We can leave it at 3 but 4 should be perfectly fine.

daq/runner.py
henry54809: The last faucet events check should only go under single shot mode so that line needs to be moved / be under another if statement. 
The reasoning for that line is it is possible for a device to finish testing before any other devices' learning events are processed considering during DAQ startup, there is a flood of events and many events are also debounced. To minimize that problem, I added in this last event processing to trigger any remaining devices to make integration tests more stable.

daq/runner.py
grafnu: THe concept there either way ("this is a set of devices that should not be tested") -- it's just a matter of what we call it.  I don't want to specifically scope it to "single shot" since that's mixing the use (single shot) with the behavior (prevent device). Would block_devices or prevent_devices be better?  What would you call something that was just meant to be a "set of devices that shouldn't be tested"?

daq/base_gateway.py
```diff
@@ -109,10 +109,12 @@ def _initialize(self):
             self.host, dummy, src_addr=self.fake_target), 'reverse ping failed'
 
     def _get_env_vars(self):
+        def opt_param(key):
+            return params.get(key) or ''  # Substitute empty string for None
         env_vars = []
         if self._env_params:
             for key, value in self._env_params.items():
-                env_vars.append(f'{key.upper()}={value}')
+                env_vars.append(f'{key.upper()}={opt_param(value)}')
```
henry54809, line 117: I this can just be `value or ''`
since the key for exists here.

henry54809 (approved):

PR #902, opened 2021-08-09
Adding discovery UDMI message

grafnu:


testing/unit/test_udmi_manager.py
```diff
@@ -0,0 +1,54 @@
...
+
+class UdmiEmptyConfigTest(UdmiDiscoverBase):
+    """Base test for empty configiration"""
+
+    BASE_CONFIG = copy.deepcopy(UdmiDiscoverBase.BASE_CONFIG)
+    del BASE_CONFIG['cloud_config']['project_id']
+
+    def test_no_manager(self):
+        """Test the case when there is an empty configuraiton"""
+        self.assertIsNone(self._mqtt)
```
henry54809, line 54: Can there also be tests for calling "discovery" and "report" when project_id is not defined?

grafnu: Done. PTAL.

PR #906, opened 2021-08-12
Fix miniet running out of IP...

henry54809:
- Fix mininet running out of IP.
- Set of controller bind to localhost only
- Clean up failed initialization in session server
- Restore using "git checkout" instead of "git switch"

### DIFF
--- a/cmd/faucet
+++ b/cmd/faucet
@@ -111,7 +111,7 @@ if [[ -n $native ]]; then
     export WSAPI_LISTEN_PORT=4567
     echo Launching $BASECMD $RYU_LISTEN_PORT $FAUCET_PROMETHEUS_PORT $WSAPI_LISTEN_PORT
     env | fgrep ${PREFIX}_
-    $BASECMD --ryu-config-file=$RYU_CONF --ryu-ofp-tcp-listen-port=$RYU_LISTEN_PORT &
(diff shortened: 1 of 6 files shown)


cmd/faucet
```diff
@@ -111,7 +111,7 @@ if [[ -n $native ]]; then
     export WSAPI_LISTEN_PORT=4567
     echo Launching $BASECMD $RYU_LISTEN_PORT $FAUCET_PROMETHEUS_PORT $WSAPI_LISTEN_PORT
     env | fgrep ${PREFIX}_
-    $BASECMD --ryu-config-file=$RYU_CONF --ryu-ofp-tcp-listen-port=$RYU_LISTEN_PORT &
+    $BASECMD --ryu-config-file=$RYU_CONF --ryu-ofp-tcp-listen-port=$RYU_LISTEN_PORT --ryu-wsapi-port=$WSAPI_LISTEN_PORT --ryu-wsapi-host=localhost --ryu-ofp-listen-host=localhost &
```
grafnu, line 114: Hmmm -- I think we need to be careful with this and/or make sure it works (testing) with some installs. If we have a provisional version (branch?) we can shoot it off to ARUP to test (although might take a few days).  My concern is the DAQ setup that has a physical switch, since in that case the OFP might be on a non-localhost connection. If so, then the easiest (I think) solution would be to make it configurable...

henry54809: That shouldn’t be affected. This change only affects the “native” faucet startup.

grafnu (approved):

PR #910, opened 2021-08-19
Support multi vxlan tunnels

henry54809:


daq/network.py
```diff
@@ -267,22 +279,37 @@ def activate(self, native_gateway=None):
...
-        if 'ip' not in vxlan_config:
-            return False
-        remote_ip = remote.ip
-        vxlan_vni = vxlan_config.get('vni', self._DEFAULT_VXLAN_VNI)
-        dst_port = int(vxlan_config.get('port', self._DEFAULT_VXLAN_PORT))
-        vxlan_cmd = self._VXLAN_CMD_FMT % (self.ext_intf, vxlan_vni, remote_ip,
-                                           dst_port, dst_port, dst_port)
+        vxlan_port = self.topology.VXLAN_SEC_TRUNK_PORT + 1
+        while vxlan_port in self._vxlan_port_sets:
+            vxlan_port += 1
```
grafnu, line 290: is there an upper limit here?  What prevents this from getting to big?

daq/session_server.py
```diff
@@ -105,9 +106,16 @@ def _init_session(self, request):
             LOGGER.info('New session stream for %s %s/%s',
                         device_mac, request.device_vlan, request.assigned_vlan)
             assert device_mac not in self._clients, 'already registered %s' % device_mac
+            vni_set = self._vni_allocations.setdefault(request.endpoint.ip, set())
+            # vni should be unique across devices in a given remote ip.
+            vni = 1
+            while vni in vni_set:
+                vni += 1
```
grafnu, line 113: is there anything that prevent this from getting too large?

daq/network.py
henry54809: Theoretically there is no limit. When a device finishes testing, the port will be freed up for use by another device. Since the vxlan tunnel is maintained even if the device is queued for testing, a port needs to be allocated immediately.

daq/session_server.py
henry54809: Same here.

grafnu:
Oh right -- this isn't a switch-port, but rather a socket-port. So, I
suppose the issue would be if the incremented port number somehow runs into
a port that is being used by something else?

henry54809:
That is allocating a switch port on the “sec” switch to setup a vxlan interface for each device. there’s a huge limit (like 20K) on the number of ovs port on a switch so I don’t see we can ever hit that.

grafnu:
Got it -- that makes sense. I hate the word "port" just b/c it has too many
meanings.

grafnu (approved):

PR #911, opened 2021-08-24
Replace objectionable terms in DAQ

anurag6:


grafnu (approved):

daq/base_gateway.py
```diff
@@ -21,7 +21,7 @@ class BaseGateway(ABC):
     """Gateway collection class for managing testing services"""
 
     GATEWAY_OFFSET = 0
-    DUMMY_OFFSET = 1
+    FAKE_OFFSET = 1
```
henry54809, line 24: I think FAKE_HOST_OFFSET is clearer. It’s not the “offset” that’s fake.

daq/network.py
```diff
@@ -29,8 +29,8 @@ class DAQHost(mininet_node.Host):
     """Base Mininet Host class, for Mininet-based tests."""
 
 
-class DummyNode:
-    """Dummy node used to handle shadow devices"""
+class PhonyNode:
```
henry54809, line 32: Why not use fake here as well?

docs/run_log.md
```diff
@@ -134,12 +134,12 @@ MM dd hh:mm:ss runner   INFO    Port 3 dpid 2 learned 9a:02:57:1e:8f:03
 MM dd hh:mm:ss runner   INFO    Gateway for device group 9a02571e8f03 not found, initializing base 3...
 MM dd hh:mm:ss gateway  INFO    Initializing gateway 9a02571e8f03 as gw03/30
 MM dd hh:mm:ss gateway  INFO    Added networking host gw03 on port 30 at 10.20.0.1
-MM dd hh:mm:ss gateway  INFO    Added dummy target dummy03 on port 31 at 10.20.0.2
-MM dd hh:mm:ss runner   INFO    Test ping gw03->dummy03
-MM dd hh:mm:ss runner   INFO    Test ping gw03->dummy03
-MM dd hh:mm:ss runner   INFO    Test ping dummy03->gw03
-MM dd hh:mm:ss runner   INFO    Test ping dummy03->192.168.84.3
-MM dd hh:mm:ss runner   INFO    Test ping gw03->dummy03 from 192.168.84.3
+MM dd hh:mm:ss gateway  INFO    Added phony target phony03 on port 31 at 10.20.0.2
```
henry54809, line 137: It should be pinging the fake host name.

daq/network.py
anurag6: I missed these files. Changed.

henry54809 (approved):

PR #917, opened 2021-08-31
Allow more devices per gateway setup

grafnu:


bin/setup_dev
```diff
@@ -147,25 +112,26 @@ if [ -n "$SIMPLE" ]; then
...
+git clone https://github.com/pseudomuto/protoc-gen-doc.git
+
 bin/setup_remote faucet
 bin/setup_remote forch
 bin/setup_remote udmi
 
-echo -n "DAQ commit "
-git log -n 1 --pretty=format:"%h - %an, %ar : %s" || true
-echo
+bin/setup_mininet
```
henry54809, line 122: i think mininet is also needed for the "simple" setup mode

daq/base_gateway.py
```diff
@@ -45,6 +44,10 @@ def __init__(self, runner, name, port_set, env_params=None):
         self.activated = False
         self.result_linger = False
         self.dhcp_monitor = None
+        is_native = bool(self.runner.run_trigger.get('native_vlan'))
+        max_hosts = self.runner.run_trigger.get('max_hosts') or DEFAULT_TEST_HOSTS
+        num_host_ports = max_hosts if is_native else DEFAULT_TEST_HOSTS
```
henry54809, line 49: it looks like theres nothing stopping a user to input a huge value here (like 1000) which will eat into the mirror port.

daq/runner.py
```diff
@@ -814,6 +816,9 @@ def _create_gateway(self, device):
             gateway.set_tap_intf(self.network.tap_intf)
 
         try:
+            if device:
+                device.gateway = gateway
+                self.network.register_device(device)
```
henry54809, line 821: I don't think the "register_device" name is appropriate here. This function is really for setting the network up for gateway (which should be device independent). I think rename to self.network.allocate_gateway_ports (or something like that) and pass in the port_set is all that's needed here. Then you can revert line 688, 794-795 and 819-821.

grafnu:
So, the underlying code (in topology) needs the device, not just the
gateway and port set... It maintains a data structure that links devices
and gateways, and therefore ports, so it's not sufficient to pass just the
port_set, it needs the device (and device.gateway) to function properly
without *major* refactoring.  So, basically the gateway ports only exist if
there is a device that references that gateway. There's no provision for a
"gateway without a device" It comes down to this function in topology.py:

    def register_device(self, device):
      """Register a device with the network topology"""
      LOGGER.info('Registering device %s port_set %s', device,
device.gateway.port_set)
      self._populate_set_devices(device, device.gateway.port_set)
      self._ensure_gateway_interfaces(device.gateway.port_set)

... which, as you see, needs both device and device.gateway.port_set

As for the name. I was intending the function to be one level up from that
specifics about what it's doing now (setting up the network ports), which
is why I went with registering.  It's not necessarily allocating anything,
as such, and I don't think it should be specific to ports.  Basically, it's
trying to say "prepare yourself for this device to do something" and then
do all the necessary parts of that (which includes the ports, but might
include other stuff). I could imagine changing register_device() to
prepare_device()
or prepare_for_device() if those make more sense?

henry54809:
The "direct_device_traffic" or "direct_port_traffic" in target_set_activate essentially does the same thing, which is informing the network of this device and sets up necessary ports/ acls. There can be some consolidations so the network setups per device aren't in 2 places.

daq/base_gateway.py
grafnu: I added an assert elsewhere. Turns out there's no easy way to do this given the current structure of the code, since the necessary values are split between topology and gateway. What I did will throw an assertion when the error condition is reached... not ideal, but better than silently just doing something weird.

grafnu:
Ok yes -- good call... updated to use direct_device_traffic

bin/setup_dev
grafnu: moved

daq/network.py
```diff
@@ -343,12 +343,13 @@ def _generate_behavioral_config(self):
             LOGGER.info('Waiting %ds for network to settle', self._settle_sec)
             time.sleep(self._settle_sec)
 
-    def direct_device_traffic(self, device, port_set):
+    def direct_device_traffic(self, device):
```
henry54809, line 346: I think this is gonna break FOT. By setting the arg port_set to None, the function cleans up the remote_tap.

grafnu: Ok, fixed.  Verified with the new fot tests that check for this case as well.

grafnu:
block_time is so that a chatty device doesn't get tested more frequently
than a non-chatty device.

block_existing is to have an on-disk mechanism for doing device scans, so
the process can be restarted without losing all state. Basically, it means
"only test devices for which there is not already a result sitting on the
disk somewhere"  The problem was with S2 with ~1000 devices... I wanted to
be able to run the system for a bit, and then possibly restart it, and over
time form a complete version of all the devices there. Without this, it
wouldn't have (device) coverage since it would keep testing the same
devices each restart.

henry54809:
block_existing won't work as intended considering everything in the run directory is cleared every time DAQ is restarted under _cleanup_previous_runs. I think a configuration option for "clear existing" and "blocked list" to not test ones that are fully tested offers more control. Block_existing may also block devices with a run directory but aren't actually tested.

grafnu:
_cleanup_previous_runs cleans out inst/reports, this is using inst/run-* as
a key for blocking runs.  There is no "run directory," as such (there is a
reports/ directory, and there are run-*/ directories, but no singular run/).

I don't know what you mean by "clear existing" and "blocked list"... Right
now, the configuration and implementation work relatively straightforward
and are sufficient for what I'm looking for. Specifically, "run the scan if
run-MACADDR does not exist" -- what additional control are you looking for
or proposing?  What actual behavior or use-case are you describing?

Note that this is NOT meant to be the full general-purpose solution, since
that case will have more cloud-centric control and repository. This is just
meant to be something that allows a CLI operator to do the needful.

henry54809:
It also clears out run-* directories. https://github.com/faucetsdn/daq/blob/master/daq/runner.py#L248

grafnu:
That's odd... because empirically it was _not_ cleaning the directories.
I'll investigate, but something is not WAI here.

Regardless of that -- what behavior were you suggesting?  The short
solution to this problem is that the system wouldn't clean out run-* if
that flag is set...

daq/base_gateway.py
```diff
@@ -147,14 +150,14 @@ def release_test_port(self, test_port):
...
-        return self.port_set * self.SET_SPACING + offset
+        return self.port_set * self._gw_set_size + offset
 
     def _is_target_expected(self, target):
         if not target:
             return False
         target_mac = target['mac']
         if target_mac in self.targets:
-            return True
+            return target_mac
```
anurag6, line 160: Looking at the code, it is not immediately clear to me why this change is needed. Based on hoqw it is used, it looks like it does expect it to be a bool.

grafnu:
Ah yes, that was a hold-over from some debugging code that I had at one
point. Removed.

henry54809:
I think by adding a run_retention and make the block_time work across restarts can achieve your intended behavior. block_existing isn't clear on what existing thing DAQ is looking for and what is being blocked. Setting block_existing also changes whether to delete existing run data is implicit and can cause confusion.

config/system/ata.yaml
```diff
@@ -19,7 +22,10 @@ switch_setup:
 
 run_trigger:
   native_vlan: 122
-  max_hosts: 3
+  max_hosts: 10
+  block_time: 10
```
henry54809, line 26: I think this and scan_interval should have a _sec postfix to match existing default_timeout_sec long_dhcp_response_sec etc. Also could there be another name for scan_interval? The name is very close to monitor_scan_sec.

grafnu:
Ok, renamed parameters and adjusted behavior!

henry54809 (approved):
Just a question about the usage of block time in non-single shot mode, everything looks good to go.

daq/runner.py
```diff
@@ -652,9 +684,11 @@ def _target_set_trigger(self, device, remote_trigger=False):
             LOGGER.info('Target device %s queing activate (%s)',
                         device, len(self._target_set_queue))
 
+        self._one_test_started = True
         if self._single_shot:
-            self._blocked_devices.add(device.mac)
-            LOGGER.info('Target device %s in now blocked (%s)', device, len(self._blocked_devices))
+            block_sec = self.run_trigger.get('device_block_sec') or LONG_TIME_SEC
+            device.set_block(block_sec)
```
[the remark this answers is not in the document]
grafnu: fixed

daq/base_gateway.py
```diff
@@ -252,38 +256,53 @@ def process_line(line):
...
         LOGGER.info('Host discovery for %s completed. Found no ip.', mac)
-        callback(None)
+        scan_callback(None)
 
-    def discover_host(self, mac: str, subnets: List[ip_network], callback: Callable):
+    def discover_host(self, mac: str, all_subnets: List[ip_network], host_callback: Callable):
         """Discovers a host using arp-scan in a list of subnets."""
         cmd = 'arp-scan --retry=2 --bandwidth=512K --interface=%s --destaddr=%s -s %s %s'
         host, intf = self._get_scan_interface()
+        scans = 5
```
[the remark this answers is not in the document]
grafnu: I updated it to make it configurable. I'm actually not sure what the right value is, so setting it down to 2 by default, and then easy to bump-up later in config if necessary. Also now possible to disable!

grafnu:
Fixed/addressed both comments (making arp_scan_count configurable).

PR #927, opened 2021-09-29
Pythonify shunt setup methods and integrate into tests

anurag6:


shunt/shell_command_helper.py
```diff
@@ -0,0 +1,96 @@
...
+        jobs = []
+        for arg_tuple in target_args:
+            process = multiprocessing.Process(target=target, args=arg_tuple)
+            jobs.append(process)
+
+        batch_start = 0
+        while batch_start <= len(jobs):
+            batch_end = batch_start + batch_size
+            if batch_end > len(jobs):
+                batch_jobs = jobs[batch_start:]
```
henry54809, line 85: this if is unnecessary. Python will slice the array correctly if batch_end > array length

anurag6: MAkes sense. Changed.

shunt/shell_command_helper.py
```diff
@@ -0,0 +1,94 @@
...
+                capture=False, docker_container=None, detach=False):
+        """
+        Args:
+            cmd: command to be executed
+            arglist: Additional arguments to add to command
+            strict: Raises exception if command execution fails
+            capture: Prints out and captures stdout and stderr output of command
+            docker_container: Set if command needs to be run within a docker container
+        returns:
+            return code of command, stdout, stderr
```
henry54809, line 52: Are the return values used anywhere? It looks like everywhere that calls run_cmd discard the return values.

henry54809 (approved):

shunt/shell_command_helper.py
anurag6: None so far, but it could be used for debugging/checking if the command went through. Bound to be useful in future dev work. It is derived from the ShellCommandHelper in Forch.

shunt/vxlan_over_ssh.py
```diff
@@ -0,0 +1,145 @@
...
+        "sudo ip link add vxlan type vxlan id %s remote %s dstport %s srcport %s %s nolearning"
+        % (vni, remote_ip, vxlan_port, vxlan_port, vxlan_port))
+    shellcmd.run_cmd("sudo ip addr add %s/24 dev vxlan" % (vtep_ip))
+    shellcmd.run_cmd("sudo ip link set vxlan up")
+
+
+def build_vxlan_ssh_conn(conn_params):
+    """
+    Args:
+        conn_params: JSON. Format:{
```
henry54809, line 123: Not important but I think this can be a dataclass instead.

anurag6: I was planning to turn it into a proto in the future. Will add a TODO.

PR #932, opened 2021-11-10
Add configuration option for nmap http tests

grafnu:


subset/pentests/nmap/test_nmap
```diff
@@ -140,7 +140,17 @@ grep -oh '[0-9]*/open[^[:space:]]*' $LOG_HTTP | while IFS=/ read -ra parts; do
     fi
 done
 
-if [ -f $TEST_ROOT/.fail ]; then
+if [[ -f $CONFIG ]]; then
```
henry54809, line 143: But it looks like nmap is still running twice regardless of whether the config is there.

subset/pentests/nmap/test_nmap
```diff
@@ -140,7 +140,17 @@ grep -oh '[0-9]*/open[^[:space:]]*' $LOG_HTTP | while IFS=/ read -ra parts; do
...
+    test_http=$(jq ".modules.pentests.test_http" $CONFIG)
```
henry54809, line 144: And why not put the configuration under the existing nmap module block?

henry54809 (approved):

PR #933, opened 2021-11-16
Refactor DTS dockerfile and startup script to be used to run DAQ in a…

anurag6:
… container locally


anurag6:
I left the dts/ files that I edited in place so that it is easier to see the changes made. I will remove dts/Dockerfile.dts and dts/start_service once I test and finalize the PR.

bin/build_daq
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash -e
+
+tag=
+
+if [ "$1" == "dts" ]; then
+  tag="dts"
```
grafnu, line 6: add "shift" just for completeness

bin/build_daq
```diff
@@ -0,0 +1,20 @@
...
+fi
+
+if [ -z "$tag" ]; then
```
grafnu, line 9: I would default tag to daq at the top, rather than a replace-if-empty conditional

bin/start_daq_container
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash -x
```
grafnu, line 1: -e

docker/modules/Dockerfile.daq
```diff
@@ -0,0 +1,42 @@
...
+COPY forch/ forch/
+
+COPY etc/ etc/
+RUN bin/setup_dev simple && rm -rf daq/.git
+
+COPY daq/ daq/
+COPY cmd/ cmd/
+COPY config/ config/
+COPY resources/ resources/
+#COPY docker/include/bin/start_daq bin/
```
grafnu, line 38: #?

grafnu (commented):
Mostly looks good -- the structure seems fine...

bin/build_daq
henry54809, line 9: this is not really an image tag; it's part of the image name.

bin/build_daq
anurag6: Done

bin/build_daq
anurag6: Done

bin/start_daq_container
anurag6: Done

docker/modules/Dockerfile.daq
anurag6: Wasn't using the newer script yet. Changed.

bin/build_daq
anurag6: Changed to image_name

bin/build_daq
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash -e
+
+image_name="daq"
+
+if [ "$1" == "dts" ]; then
+  image_name="dts"
+  shift
+fi
+
+bin/setup_remote faucet
```
grafnu, line 10: Script is misnamed since it's not "building daq" as such... so maybe "build_daq_container"?  (dts was specific to it being a container, daq is not so qualified).

grafnu (approved):
Just one file to rename, and then good to go!

bin/build_daq
anurag6: Makes sense. Canged.

PR #934, opened 2021-11-29
Refactor ShellCommandHelper and add OvsHelper

anurag6:
The PR adds a module that helps create the OVS n/w setup needed to query DAQ interface/simulate an access switch.


bin/build_daq_container
```diff
@@ -12,6 +12,6 @@ bin/setup_remote forch
 
 echo Starting $image_name build...
 
-docker build . -f docker/modules/Dockerfile.daq --build-arg GIT_URL -t daqf/$image_name
+docker build . -f docker/modules/Dockerfile.daq -t daqf/$image_name
```
grafnu, line 15: why remove this?

device_coupler/ovs_helper.py
```diff
@@ -0,0 +1,79 @@
...
+
+from __future__ import absolute_import, division
+
+from functools import partial
+from python_lib.shell_command_helper import ShellCommandHelper
+from utils import get_logger
+
+
+class OvsHelper:
+    """Class to build OVS, VxLANs and other network components"""
```
grafnu, line 11: I don't think "build OVS" is what you mean here -- that implies actually compiling OVS (which is a thing). Do you mean something like construct an OVS bridge?

device_coupler/ovs_helper.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def create_ovs_bridge(self, name):
+        """Creates OVS bridge"""
+        self._logger.info('Creating OVS bridge %s', name)
+        self._run_shell('sudo ovs-vsctl add-br %s' % name)
+
+    def delete_ovs_bridge(self, name):
+        """Delete ovs bridge"""
+        self._logger.info('Deleting OVS bridge %s', name)
+        self._run_shell_no_raise('sudo ovs-vsctl add-br %s' % name)
```
henry54809, line 49: shouldn't this be for del br

bin/build_daq_container
anurag6: I thought that arg wasn't getting consumed anywhere. Is that a mistake?

device_coupler/ovs_helper.py
anurag6: Yup. Updated.

device_coupler/ovs_helper.py
anurag6: I meant OVS network components but updated it.

bin/build_daq_container
grafnu: I forget exactly how it works, but it sets an environment variable which is consumed somewhere down the chain of things. Docs seem to imply that it also requires an ENV setting in the Dockerfile, so not sure if it works without that or if it's broken.

grafnu (approved):

bin/build_daq_container
anurag6: Fair enough. I reverted that particular bit. Thought I was cleaning up something. Better not to break things.

device_coupler/ovs_helper.py
```diff
@@ -0,0 +1,79 @@
...
+        self._logger.info('Enabling native VLAN %s on interface %s', vlan, interface)
+        self._run_shell('sudo ovs-vsctl set port %s tag=%s' % (interface, vlan))
+
+    def set_trunk_vlan(self, interface, vlans):
+        """Takes an array of VLANs and sets them as trunk VLANs for the port on OVS bridge"""
+        self._logger.info('Enabling trunk VLANs %s on interface %s', vlans, interface)
+        vlan_str = ",".join(str(vlan) for vlan in vlans)
+        self._run_shell('sudo ovs-vsctl set port %s trunks=%s' % (interface, vlan_str))
+
+    def create_faux_device(self, index):
```
henry54809, line 67: Are we going to refactor cmd/faux to python?

anurag6: Probably a good idea long term, but I am not working on it atm. I just need the faux as a testing tool and simply running it as a shell script works for me.

PR #940, opened 2021-12-19
Make 802.1x test easier to use

noursaidi:
- Adds more meaningful messages and logging
- Move certificates to a more user-friendly location


subset/dot1x/authenticator/authenticator.py
```diff
@@ -290,17 +290,20 @@ def run_authentication_test(self):
...
             test_result = "pass"
             for src_mac, is_success in self.results.items():
+                additional = ''
                 if is_success:
                     result = 'succeeded'
                 else:
                     result = 'failed'
                     test_result = "fail"
-                result_str += "Authentication for %s %s." % (src_mac, result)
+                    additional = ' Incorrect credentials provided.'
```
anurag6, line 305: I think incorrect credentials isn't the only reason for a test failure. Adding that as a reason for every failure might be misleading

noursaidi: I've changed it so incorrect credentials only appears when incorrect credentials is the reason by checking if there is a RadiusAccessReject which suggests the EAP process was fine and the credentials were incorrect.

PR #942, opened 2021-12-24
Add event system and config parsing to device coupler 

anurag6:


proto/system_config.proto
```diff
@@ -315,4 +315,28 @@ message CloudConfig {
...
+ */
+message DeviceCouplerConfig {
+  // Name of trunk interface that connects to access switch
+  string trunk_iface = 1;
+
+  // Name of OVS bridge where device MACs from access s/w are learnt
+  string bridge = 2;
+
+  // gRPC target IP
+  string target_ip = 3;
```
henry54809, line 331: Trevor will probably say "target" is ambiguous.

device_coupler/device_discovery.py
```diff
@@ -0,0 +1,76 @@
...
+        event = DeviceDiscoveryEvent()
+        event.port = int(port)
+        event.vlan = int(vlan)
+        event.mac = mac
+        event.event_type = DiscoveryEventType.EXPIRY if expire else DiscoveryEventType.DISCOVERY
+        return event
+
+
+def main():
+    pass
```
henry54809, line 72: What's this for if it's not intended to be run as a script?

device_coupler/ovs_helper.py
```diff
@@ -95,3 +97,29 @@ def create_veth_pair(self, iface1, iface2):
...
+        """Converts fdb  table string to list[(<port>, <vlan>, <mac>)]"""
+        filtered_table = [entry for entry in table.split('\n') if self._is_mac_in_string(entry)]
+        return [tuple(entry.split()[:-1]) for entry in filtered_table]
+
+    def get_interface_ip(self, iface="eth0"):
+        """Returns IP of fiven interface"""
+        _, cmd_str, _ = self._run_shell('ip addr show %s' % iface)
+        _, _, ip_str = cmd_str.partition('inet')
+        ip = ip_str.split()[0].split('/')[0] if ip_str else None
+        return ip
```
henry54809, line 125: And I think this file is easily unit testable

device_coupler/runner.py
```diff
@@ -0,0 +1,110 @@
...
+                event = self._event_queue.get(timeout=self._WORKER_TIMEOUT)
+                self._logger.info(event)
+                if event.event_type == DiscoveryEventType.DISCOVERY:
+                    self._daq_client.process_device_discovery(event.mac, event.vlan)
+                else:
+                    self._daq_client.process_device_expiry(event.mac)
+            except Empty:
+                # Worker thread timeout. Do nothing
+                self._logger.info('Anurag worker timeout')
+                pass
```
henry54809, line 91: pass is not needed since there's already another line under except

device_coupler/runner.py
```diff
@@ -0,0 +1,110 @@
...
+        """Setup device coupler"""
+        self._running = True
+        self._ovs_helper = OvsHelper()
+
+        self._network_helper = NetworkHelper(
+            self._config.trunk_iface, self._config.bridge, self._test_vlans)
+        self._network_helper.setup()
+
+        self._event_queue = Queue()
+        self._build_worker_threads(self._process_event_queue, self._WORKER_COUNT)
```
henry54809, line 48: I think you can simplify these by using a ThreadPoolExecutor here for process_event_queue and polling of the forwarding table.

device_coupler/ovs_helper.py
```diff
@@ -95,3 +97,29 @@ def create_veth_pair(self, iface1, iface2):
...
+        """Returns true if string contains a mac"""
+        return bool(re.search("[0-9a-f]{2}([:]?)[0-9a-f]{2}(\\1[0-9a-f]{2}){4}", string.lower()))
+
+    def _filter_forwarding_table(self, table):
+        """Converts fdb  table string to list[(<port>, <vlan>, <mac>)]"""
+        filtered_table = [entry for entry in table.split('\n') if self._is_mac_in_string(entry)]
+        return [tuple(entry.split()[:-1]) for entry in filtered_table]
+
+    def get_interface_ip(self, iface="eth0"):
+        """Returns IP of fiven interface"""
```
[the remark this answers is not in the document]
anurag6: Fixed.

device_coupler/ovs_helper.py
```diff
@@ -95,3 +97,29 @@ def create_veth_pair(self, iface1, iface2):
...
+        """Returns forwarding table of given OVS bridge in the form [(<port>, <vlan>, <mac>)]"""
+        _, out, _ = self._run_shell('ovs-appctl fdb/show %s' % bridge)
+        return self._filter_forwarding_table(out)
+
+    def _is_mac_in_string(self, string):
+        """Returns true if string contains a mac"""
+        return bool(re.search("[0-9a-f]{2}([:]?)[0-9a-f]{2}(\\1[0-9a-f]{2}){4}", string.lower()))
+
+    def _filter_forwarding_table(self, table):
+        """Converts fdb  table string to list[(<port>, <vlan>, <mac>)]"""
```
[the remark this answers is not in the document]
anurag6: Fixed.

device_coupler/ovs_helper.py
```diff
@@ -95,3 +97,29 @@ def create_veth_pair(self, iface1, iface2):
...
+        filtered_table = [entry for entry in table.split('\n') if self._is_mac_in_string(entry)]
+        return [tuple(entry.split()[:-1]) for entry in filtered_table]
+
+    def get_interface_ip(self, iface="eth0"):
+        """Returns IP of fiven interface"""
+        _, cmd_str, _ = self._run_shell('ip addr show %s' % iface)
+        _, _, ip_str = cmd_str.partition('inet')
```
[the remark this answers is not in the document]
anurag6: Makes sense. Changed.

device_coupler/ovs_helper.py
```diff
@@ -95,3 +97,29 @@ def create_veth_pair(self, iface1, iface2):
...
+        return self._filter_forwarding_table(out)
+
+    def _is_mac_in_string(self, string):
+        """Returns true if string contains a mac"""
+        return bool(re.search("[0-9a-f]{2}([:]?)[0-9a-f]{2}(\\1[0-9a-f]{2}){4}", string.lower()))
+
+    def _filter_forwarding_table(self, table):
+        """Converts fdb  table string to list[(<port>, <vlan>, <mac>)]"""
+        filtered_table = [entry for entry in table.split('\n') if self._is_mac_in_string(entry)]
+        return [tuple(entry.split()[:-1]) for entry in filtered_table]
```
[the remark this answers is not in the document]
anurag6: Changed to use regex

proto/system_config.proto
anurag6: Changed to daq_grpc_ip

device_coupler/device_discovery.py
anurag6: Testing remnants. Cleared.

device_coupler/ovs_helper.py
anurag6: Added a unit test file for ovs_helper.

device_coupler/runner.py
anurag6: The log crept in which shouldnt be there. Removed it. Retaining the pass.

device_coupler/runner.py
anurag6: I figured ThreadPoolExecutor might be overkill for what I Was doing. Replaced it with ThreadPoolExecutor though.

device_coupler/ovs_helper.py
```diff
@@ -80,18 +87,28 @@ def set_trunk_vlan(self, interface, vlans):
...
+
+    def _filter_forwarding_table(self, table):
+        """Converts fdb table string to list[(<port>, <vlan>, <mac>)]"""
+        filtered_table = re.findall(r"\d{1,4}\s+\d{1,4}\s+(?:[0-9a-fA-F]:?){12}", table)
+        return [tuple(entry.split()) for entry in filtered_table]
+
+    def get_interface_ip(self, iface="eth0"):
+        """Returns IP of given interface"""
+        _, cmd_str, _ = self._run_shell('ip addr show %s' % iface)
+        return re.search(r'(\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3})', cmd_str).group(0)
```
henry54809, line 114: Should there be some error handling in case there is no interface ip?

henry54809 (commented):
I've also added trevor as a reviewer so he can review them tomorrow.

device_coupler/ovs_helper.py
anurag6: I think this should fail catastrophically. The burden of error handling should be on DAQ or those using these methods. I think this comes into play when device coupler is integrated with DAQ.

henry54809: I mean for error handling it can raise a sensible error here rather than “None type has no attribute group”.

device_coupler/runner.py
```diff
@@ -0,0 +1,109 @@
...
+                self._logger.info(event)
+                if event.event_type == DiscoveryEventType.DISCOVERY:
+                    self._daq_client.process_device_discovery(event.mac, event.vlan)
+                else:
+                    self._daq_client.process_device_expiry(event.mac)
+            except Empty:
+                # Worker timeout. Do nothing
+                pass
+
+    def _build_worker_threads(self, method, count):
```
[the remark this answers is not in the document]
anurag6: Whoops. Changed.

device_coupler/ovs_helper.py
anurag6: Makes sense.

anurag6: Changed to add error handling. I think rum_cmd should handle it well enough but this prevents dumping the whole stack trace.

.github/workflows/tests.yml
```diff
@@ -244,13 +248,14 @@ jobs:
...
       run: |
-        bin/load_images faucet faux setup base
+        bin/load_images faucet faux setup base subset
     - name: Running device_coupler test
       env:
         DOCKER_STARTUP_TIMEOUT_MS: 60000
         GCP_BASE64_CRED: ${{ secrets.GCP_BASE64_CRED }}
         GCP_REFLECT_KEY_BASE64: ${{ secrets.GCP_REFLECT_KEY_BASE64 }}
       run: |
+        python3 -m unittest discover -s device_coupler/testing/unit -p "test_*.py"
```
grafnu, line 258: unless there's a really good reason for it, all the testing stuff should be under testing/ (not device_coupler/testing) -- then this line itself shouldn't be necessary since it's covered under the regular slate of unit tests.

daq/topology.py
```diff
@@ -429,11 +430,18 @@ def _generate_switch_local_acls(self, portset_acls, local_acl):
...
+            for device in devices:
+                if device and device.gateway:
+                    target = {'port': None, 'port_set': device.gateway.port_set, 'mac': device.mac}
+                    targets.append(target)
+        for target in targets:
             port_no = target['port']
             port_set = target['port_set']
             target_mac = target['mac']
-            mirror_port = self.mirror_port(port_no)
+            # Mirror port is set to tap_intf, which is pri trunk port if no target port
```
grafnu, line 443: I don't understand this comment. It's not explaining "why" and/or it doesn't match what the code says.

daq/topology.py
```diff
@@ -488,14 +496,23 @@ def _add_dot1x_incoming_rule(self, incoming_acl, secondary_acl):
...
                         self._add_dot1x_allow_rule(incoming_acl, test_ports, vlan_vid=vlan)
                     device_port = device.port.port_no
                     if device_port:
                         self._add_dot1x_allow_rule(secondary_acl, [device_port], in_vlan=vlan)
 
+    def _add_dot1x_vxlan_acl(self, acls):
+        for devices in self._set_devices.values():
+            for device in devices:
+                if device and device.vlan:
+                    acl_name = 'vxlan_coupler_%s' % device.vlan
```
grafnu, line 511: something seems mismatched here since the function name is vxlan_acl but this is adding a coupler-specific acl?

daq/topology.py
```diff
@@ -535,9 +552,14 @@ def _generate_main_acls(self):
...
         acls[self.INCOMING_ACL_FORMAT % self._COUPLER_ACL] = [
             self._make_acl_rule(ports=[self.VXLAN_SEC_TRUNK_PORT], allow=True)]
+        self._add_dot1x_vxlan_acl(acls)
 
         for port_set in range(1, self.sec_port):
             vlan = self._port_set_vlan(port_set)
+            if port_set in self._set_devices:
+                portset_device = next(iter(self._set_devices[port_set]))
+                if portset_device.vlan:
+                    vlan = portset_device.vlan
```
grafnu, line 562: try to avoid mutable variables with any significant scope. You can/should rewrite this such that you calculate an intermediate variable and then assign to vlan once (vlan = portset_device_vlan or self._port_set_vlan), or better yet can this logic be folded into the function (_port_set_vlan)?

device_coupler/config/device_coupler_config.yaml
```diff
@@ -0,0 +1,6 @@
+trunk_iface: trunk0
```
grafnu, line 1: as discussed, I think some of these are redundant with existing parameters (test_vlans, grpc_port), and then I'd find the right way to fold the others into the existing setup. E.g. the "daq_grpc_ip" argument really should be collocated with the (existing) server_port parameter (e.g. server_ip).

device_coupler/heartbeat_scheduler.py
```diff
@@ -0,0 +1,39 @@
+"""Schedule heart beat with functions that have to be called in order"""
+from __future__ import absolute_import
+import threading
+
+from device_coupler.utils import get_logger
+
+
+class HeartbeatScheduler:
+    """Heart beat scheduler"""
```
grafnu, line 9: Heartbeat is one word (check comments here and above)

device_coupler/runner.py
```diff
@@ -0,0 +1,103 @@
...
+                else:
+                    self._daq_client.process_device_expiry(event.mac)
+            except Empty:
+                # Worker timeout. Do nothing
+                pass
+
+
+def main():
+    config = yaml_proto(
+        "device_coupler/config/device_coupler_config.yaml", sys_config.DeviceCouplerConfig)
```
grafnu, line 94: is this for testing or is this the normal flow of things?  I'm just wondering about this path used as part of normal operation outside of testing.

device_coupler/testing/test_device_coupler.out
```diff
@@ -1,7 +1,9 @@
+9a:02:57:1e:8f:01
+=================
 RESULT pass base.startup.dhcp
 RESULT skip base.switch.ping
 RESULT pass connection.base.target_ping
-RESULT pass security.nmap.ports
+RESULT fail security.nmap.ports
```
grafnu, line 6: so this doesn't seem right?  You mentioned something about this when we chatted... would be a good idea to resolve the underlying issue first if there is one.

device_coupler/runner.py
anurag6: This is just an intermediate state of things where we have device coupler runnign as a separate entity as part of a script. I think it is fair to say this is exclusively for test_device_coupler to go through. The next step would be DAQ feeding its own config params to device coupler and running it.

device_coupler/heartbeat_scheduler.py
anurag6: Changed

device_coupler/config/device_coupler_config.yaml
anurag6: Changed device coupler so that it reads. DAQ. config instead

daq/topology.py
anurag6: Added more detailed comment.

.github/workflows/tests.yml
anurag6: Relocated unit tests

daq/topology.py
anurag6: Changed method name to reflect coupler_acl instead.

daq/topology.py
anurag6: Moved it to _port_set_vlan instead.

device_coupler/testing/test_device_coupler.out
anurag6: This delta shows up because the earlier device_coupler blocked/did not route certain kinds of traffic. This also resulted in tests like ncrack and dot1x. not working. This change makes sure all tests work and takes care of that. The fail result is actually what is expected and is the same as what is seen in the output of test_aux.

device_coupler/device_discovery.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def cleanup(self):
+        """Clean up device discovery"""
+        if self._polling_timer:
+            self._polling_timer.stop()
+        self._logger.info('Clean up complete.')
+
+    def _process_entry(self, event):
+        # Only process events for test vlans in config and on trunk port
+        if event.port == self._trunk_ofport and\
```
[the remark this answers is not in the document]
anurag6: Changed

daq/network.py
```diff
@@ -72,7 +72,10 @@ def __init__(self, config):
         self._used_ip_indices = set()
         self.topology = FaucetTopology(self.config)
         switch_setup = config.get('switch_setup', {})
-        self.ext_intf = switch_setup.get('data_intf')
+        if self.config.get('run_trigger', {}).get('auto_session'):
```
grafnu, line 75: Not a big fan of having this logic duplicated. I think it would be trivial to have network.py pull it from topology... i.e.

`self.ext_intf = self.topology.ext_intf`

and then leave the if statement logic in just that one place.

device_coupler/config/daq_config.yaml
```diff
@@ -0,0 +1,27 @@
...
+
+run_trigger:
+  vlan_start: 200
+  vlan_end: 205
+  auto_session: True
+
+device_reporting:
+  server_port: 50051
+
+docker_bridge_ip: 172.17.0.1
```
grafnu, line 27: this shouldn't be "docker" -- what's it used for? (Not where does it come from.)  Is this the address that the gRPC client tries to connect to?  Also, likely should be under some subsection not at the top-level.  Is this used by the device-coupler inside of its docker container, right?  Now that I think more about it, it might be better as an ENV variable just passed in when the coupler container is started... would that be easy to do?

device_coupler/runner.py
grafnu: I think both pass and # are also not both required?

anurag6: Comments are ignored by the interpreter. Still seems to crib if there is no pass

device_coupler/config/daq_config.yaml
anurag6: I think that's should be doable, but it necessitates the integration of DAQ and device_coupler.

daq/network.py
anurag6: Changed.

device_coupler/config/daq_config.yaml
anurag6: I moved docker_bridge_ip into run_trigger instead.

grafnu: You missed the part about the name (docker_ prefix)

anurag6: Welp. Renamed to grpc_target_ip

device_coupler/config/daq_config.yaml
```diff
@@ -20,7 +20,7 @@ run_trigger:
   vlan_start: 200
   vlan_end: 205
   auto_session: True
-  docker_bridge_ip: 172.17.0.1
+  grpc_target_ip: 172.17.0.1
```
grafnu, line 23: not grpc -- not a useful prefix. Think of the semantics of what it's doing and come up with a descriptive name.  It's the IP address for some machine -- what is the purpose of that target?

grafnu (approved):

Issue #886, opened 2021-06-22
Pics Failing Unsupported Types Instead of Skipping


jhughesbiot:
If the pics file has the value false in the supported column, the pics statement is generating failure results instead of skipping. It appears there is no catch all for when a property is not supported so the default behavior is to use the false value in the testPassed variable.

This is where I believe the change needs to occur.
https://github.com/faucetsdn/daq/blob/73d7504f7264f3b3bfafb378bdf04fe87e882781/subset/bacnet/bacnetTests/src/main/java/helper/PicsValidator.java#L21

Likely not caught due to the faux devices pics.csv being used in testing has no objects defined as false in the support column.

https://github.com/faucetsdn/daq/blob/73d7504f7264f3b3bfafb378bdf04fe87e882781/docker/include/bacnet/bacnetFaux/src/main/resources/pics.csv

PR #948, opened 2022-02-14
Fix Pics Statement Failing Unsupported Properties

jhughesbiot:
Corrects issue: https://github.com/faucetsdn/daq/issues/886

Properties indicated as not supported, will be shown as SKIPPED in the result.  The same will occur if an object is in the pics statement but is not in the device as long as all properties are marked false.

Tested against the most recent device that had this failure present itself to confirm both test builds and real world scenarios.  Image included to show how the report and pics statement handle these scenarios.

image

### DIFF
--- a/docker/include/bacnet/bacnetFaux/src/main/resources/pics.csv
+++ b/docker/include/bacnet/bacnetFaux/src/main/resources/pics.csv
@@ -1,4 +1,5 @@
 Bacnet_Object_Type,Bacnet_Object_Property,Property_Datatype,Conformance_Code,Supported,
+Bacnet_Analogue_Input,Unsupported_Property,,R,FALSE,
 Bacnet_Analogue_Input,Object_Identifier,BACnetObjectIdentifier,R,TRUE,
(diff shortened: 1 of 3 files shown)


grafnu (approved):

jhughesbiot:
@grafnu 
Is there any way to get a report of the stickler-ci failure for a single file?  I'd like to get that issue resolved before merging just not seeing anyway to run the formatting check against a single file and get a report.

grafnu:
I don't know how to do that specifically (single file), but if you click
through on the detail of the failure it takes you to a summary page:
https://github.com/faucetsdn/daq/pull/948/checks?check_run_id=5192330626

Looks like there's an entry in .checkstyle.xml that's tripping it up. Not
sure what changed (likely a newer version of stickler checks...)

jhughesbiot:
I think I have a fix for the stickler but would want to get this merged in first so I can get a clean PR going to just address that issue.  I can't merge it in since I don't have write access.

PR #949, opened 2022-02-15
Fix stickler config

jhughesbiot:
Update stickler config to fix stickler-ci from failing.


grafnu:
There are no changed files?  And lots of tests are breaking?

jhughesbiot:
I'm not entirely sure why it's not detecting the changed files, you can see in the commit history the files that were changed.  All the failed tests I looked at appeared to be the test itself failing to download something and then timing out but not actually running.

jhughesbiot:
I'll close this one and create a new one that is detecting the files correctly.

PR #953, opened 2022-02-22
Target port alternative for switch tests

anurag6:


daq/host.py
```diff
@@ -103,6 +103,7 @@ def __init__(self, runner, device, config):
         self.device = device
         self.target_mac = device.mac
         self.target_port = device.port.port_no
+        self.nonof_dev_port = str(device.nonof_port_num) if device.nonof_port_num else None
```
grafnu, line 106: this logic seems wonky, and I'm guessing you're trying to take a shortcut with the definition of self.target_port.  Under what conditions is the meaning of "self.target_port" not the useful value here?  Seems like the concept of a target_port should apply to all switch cases (they all have the same sequential alignment of ports).

You mentioned that just setting self.target_port to something causes problems... but I would think of this the other way around and figure out what trigger to use for those cases rather than changing the foundational meaning of self.target_port, which is ultimately very confusing.

daq/host.py
```diff
@@ -324,15 +325,16 @@ def _state_transition(self, target, expected=None):
     def _build_switch_info(self) -> usi.SwitchInfo:
         switch_config = self._get_switch_config()
         model_str = switch_config['model']
-        if model_str == 'FAUX_SWITCH' or not self.target_port:
+        target_port = self.target_port or self.nonof_dev_port
```
grafnu, line 328: case in point -- this logic should be unnecessary.

daq/host.py
```diff
@@ -768,7 +770,7 @@ def _get_module_params(self):
             'local_ip': ext_loip,
             'target_ip': self.target_ip,
             'target_mac': self.target_mac,
-            'target_port': str(self.target_port) if self.target_port else None,
+            'target_port': str(self.target_port) if self.target_port else self.nonof_dev_port,
```
grafnu, line 773: another code smell here -- you're duplicating the logic.  If anywhere, this should go in the constructor, and keep these occurrences as self.target_port and replace OTHER uses of self.target_port with a more appropriate value that induces the relevant behavior.

daq/runner.py
```diff
@@ -523,10 +525,12 @@ def _handle_device_learn(self, target_mac, vid):
         else:
             self._target_set_trigger(device)
 
-    def _remote_trigger(self, device, device_vlan, assigned_vlan):
+    def _remote_trigger(self, device, device_vlan, assigned_vlan, device_port):
         assert device_vlan, 'expected device_vlan'
         device.port.flapping_start = 0
         device.port.active = True
+        if device_port:
```
grafnu, line 532: this doesn't seem right.  The device is on a target_port no matter what the of situation

device_coupler/daq_client.py
```diff
@@ -106,9 +107,10 @@ def _run_test_session(self, mac, session):
             self._logger.error('Traceback: %s', traceback.format_exc())
         self._disconnect(mac)
 
-    def _initiate_test_session(self, mac, device_vlan):
+    def _initiate_test_session(self, mac, device_vlan, port):
         if mac in self._mac_sessions:
-            if device_vlan == self._mac_sessions[mac]['device_vlan']:
+            if device_vlan == self._mac_sessions[mac]['device_vlan'] and\
```
grafnu, line 112: don't use line continuation. Refactor (introduce local var?) instead.

device_coupler/runner.py
```diff
@@ -84,13 +84,19 @@ def _process_event_queue(self):
...
+                    self._daq_client.process_device_discovery(event.mac, event.vlan, port)
                 else:
                     self._daq_client.process_device_expiry(event.mac)
-            except Empty:
+            except Exception:
                 # Worker timeout. Do nothing
                 pass
 
+    def _get_device_port(self, vlan):
+        """Temporary mapping for VLAN to ports uses first VLAN of accepted test VLAN range."""
```
grafnu, line 96: Why is this temporary?  Just indicate what it does without getting into how long it might exist :-)

daq/host.py
noursaidi: target_port needs to be an integer, so assigning  nonof_dev_port which is a string causes an error - 
```Feb 24 12:46:47 ipaddr.ipaddr01 ERROR   Exception for 38d135021d24: '2' has type str, but expected one of: int, long ```

Raised by line `return usi.SwitchInfo(**params)` (a protobuf) in the `_buid_switch_info` function in `host.py`

**Comment by noursaidi** [comment]
[This line which isn't part of this PR](https://github.com/faucetsdn/daq/blob/b832da4997d0af8d100b2aaecffd86ff1b71e178/device_coupler/runner.py#L97) seems to hardcode the system configuration file which is different to the `local/system.yaml` - that needs updating so it uses the system configuration file

File: daq/:91 [hunk 7]
```diff
@@ -0,0 +1,1286 @@
+"""Main test runner for DAQ"""
```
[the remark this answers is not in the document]
anurag6: Yup. Removed

daq/host.py
anurag6: Ah. Forgot to update this in the repo. Reworked the file. Obsolete now.

daq/host.py
anurag6: Changed.

device_coupler/runner.py
anurag6: Changed.

device_coupler/daq_client.py
anurag6: Changed.

daq/runner.py
anurag6: Changed.

daq/host.py
```diff
@@ -102,7 +102,8 @@ def __init__(self, runner, device, config):
         self._no_test = self.config.get('no_test', False)
         self.device = device
         self.target_mac = device.mac
-        self.target_port = device.port.port_no
+        self.target_port = device.port.port_no or device.nonof_port_num
```
grafnu, line 105: this still doesn't seem right -- why is "device.port.port_no" not always set correctly?

daq/host.py
```diff
@@ -290,7 +291,7 @@ def initialize(self):
         os.makedirs(self.scan_base)
         self._initialize_config()
         network = self.runner.network
-        if self.target_port:
+        if self.target_port and self._target_port_mirror:
```
grafnu, line 294: this should just be "self._use_target_port_mirror" and have any conditional logic where the variable is set.  Then you don't need to duplicate the logic in two places.

daq/runner.py
```diff
@@ -523,10 +525,11 @@ def _handle_device_learn(self, target_mac, vid):
         else:
             self._target_set_trigger(device)
 
-    def _remote_trigger(self, device, device_vlan, assigned_vlan):
+    def _remote_trigger(self, device, device_vlan, assigned_vlan, device_port):
         assert device_vlan, 'expected device_vlan'
         device.port.flapping_start = 0
         device.port.active = True
+        device.nonof_port_num = device_port
```
grafnu, line 532: The _device_ entry shouldn't care about OF vs. non-OF... this should just be setting the device port in all cases.

device_coupler/start_daq
```diff
@@ -6,4 +6,4 @@ ENDPOINT_CFG=/tmp/startup.yaml
 DAQ_CONFIG=device_coupler/config/daq_config.yaml
 echo "site_description: generated" > $ENDPOINT_CFG
 yq e -n ".switch_setup.endpoint.ip=\"$docker_br_ip\"" >> $ENDPOINT_CFG
-sudo PYTHONPATH=$daq_path:$daq_path/faucet:$daq_path/forch:$daq_path/mininet PATH=$PATH:$daq_path/mininet FAUCET_EVENT_SOCK=$daq_path/inst/faucet_event.sock cmd/start $DAQ_CONFIG $ENDPOINT_CFG "$@"
+sudo PYTHONPATH=$daq_path:$daq_path/faucet:$daq_path/forch:$daq_path/mininet PATH=$PATH:$daq_path/mininet FAUCET_EVENT_SOCK=$daq_path/inst/faucet_event.sock cmd/start $DAQ_CONFIG $ENDPOINT_CFG "$@" || true
```
grafnu, line 9: ??  Under what condition does this fail and you want to keep going?

proto/session_server.proto
```diff
@@ -19,6 +19,9 @@ message SessionParams {
 
   // Information about client tunnel endpoint.
   TunnelEndpoint endpoint = 4;
+
+  // Openflow port on which device was learned
```
grafnu, line 23: No need to mention that it's open-flow -- it's just the port number. And it's not "was learned" -- it's just where it *is* (the fact that it was done through learning is immaterial here).

grafnu (commented):
There's still something weird about how you're passing the device port around...  The "device port" concept is universal and should be consistent across all modes. I think the trick is figuring out how to signal the logic of what to do about things, but it's not a "device" consideration.

anurag6:
> There's still something weird about how you're passing the device port around... The "device port" concept is universal and should be consistent across all modes. I think the trick is figuring out how to signal the logic of what to do about things, but it's not a "device" consideration.

I think I disagree. In the future if device_coupler gets the capability of managing/viewing the access switch through gNMI, it would be device_coupler that knows what access port the device is on. I think it makes sense to populate it at the device_coupler level and pass it on.

grafnu:
I don't see how that changes the basic assertion: Each device has a port,
there is a field in the device structure that says what port it's on, so
when there is a known port then that field should be set correctly. The
problem is your current solution bleeds the abstraction into places it
shouldn't go, and I'm not convinced by the speculative future use case...
Maybe in the future the system doesn't know anything about the port, but
that still doesn't argue for having the port encoded in weird ways now.

device_coupler/start_daq
anurag6: TimeoutError. It happens in test_aux as well and it doesnt fail since -e isn't set. It istesting the timeout and expects it to happen.

grafnu:
Ok -- please just add a clarifying comment (just what you said here)

daq/host.py
anurag6: device.port.port_no is used to denote port triggered tests in [daq/runner.py.](https://github.com/anurag6/daq/blob/59e9ba44a37790c9fea294a2beb59229da6e4ed1/daq/runner.py#L715). Setting it there causes all sorts of issues becase of that assumption.  I am not sure how to remove that assumption without causing significant issues.

daq/runner.py
anurag6: device.port.port_no is relied upon to indicate a port trigger as mentioned in the comment above. this was a work_around around it.

daq/host.py
anurag6: Fixed

proto/session_server.proto
anurag6: Fixed.

grafnu (approved):

